PR #2, opened 2026-07-20
docs: PR #1 review notes + agent ops — the voice, the evidence, the definition of done

zaphodis42:
## What

Two docs, two artifacts, one nit fix — the outputs of a post-merge review of PR #1:

- `docs/superpowers/specs/2026-07-21-pr1-review-notes.md` — four owner-confirmed findings: GTM.md still describes the pre-fork schedule; the replacement ADR should record the honest rationale (the switching UX was judged unacceptable — not "impossible"); the crash-freedom gate needs eyes (opt-in ACRA-style reporting, decided in the fork ADR); and two licence nuances (core modules can stay Apache-2.0 inside a GPL-3.0 app; onboarding's glide-typing promise is a check the base has to cash).
- `docs/superpowers/specs/2026-07-21-agent-ops-recommendations.md` — the agent persona (the staff writes docs, guides, and conversation in the VOICE.md register, with a drafted "How you talk" AGENTS.md section and hard limits), a Definition of Done for "fully complete PRs", separation of duties + cross-family review borrowed from darkmill's ADR log, and a four-rung evidence ladder: prompt goldens → Paparazzi screenshot goldens per PR → emulator journeys with machine-checkable asserts → an mp4 per journey that doubles as the GTM demo asset.
- `PATCHNOTES.md` — the voice demonstrated on everything merged so far, rather than described.
- `.github/pull_request_template.md` — Evidence and Patch-note sections, so completeness is a checklist instead of a mood.
- Mockups README: row 10 now admits its strip is the rejected four-circle variant.

## Why

The owner asked for a review of PR #1 and for recommendations on agent maintenance, an agent persona, and evidence-driven PRs. Nothing in the recommendations is law until it lands in AGENTS.md or an ADR — this PR is the proposal, plus the two artifacts cheap enough to just ship.

## Evidence

- **Commands run + output:** docs-only change; `personas/` and prompt code untouched, so rule 4 does not trigger. Facts in PATCHNOTES.md cross-checked against `git log`.
- **Graded by:** the author, which the enclosed recommendations say should stop being allowed. The irony is noted in the minutes.

## Patch note

The robot staff reviewed its own previous PR, proposed rules that would forbid exactly that, and wrote them down in a voice it is now contractually wearing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Documentation**
  - Added a standardized pull request template with sections for changes, rationale, evidence, patch notes, and validation.
[...]

PR #3, opened 2026-07-20
docs: fork-spike results — HeliBoard vs AnySoftKeyboard vs FlorisBoard graft evidence

zaphodis42:
## What this is

The fork-base spike from [`2026-07-20-fork-spike-checkpoint.md`](https://github.com/apexcloudwise/personaspeak/blob/docs/fork-spike-results/docs/superpowers/specs/2026-07-20-fork-spike-checkpoint.md): clone, build, and graft the real persona-strip UI (per the [UX design](https://github.com/apexcloudwise/personaspeak/blob/docs/fork-spike-results/docs/superpowers/specs/2026-07-20-keyboard-ux-design.md)) onto all three candidate keyboard bases, then grade how invasive each graft actually is. Full log with per-candidate detail lives in [`docs/superpowers/specs/2026-07-20-fork-spike-results.md`](https://github.com/apexcloudwise/personaspeak/blob/docs/fork-spike-results/docs/superpowers/specs/2026-07-20-fork-spike-results.md) — this description is a summary of that file.

**This PR does not decide the base.** It's evidence for that decision, which the checkpoint calls out as the owner's to make deliberately (it's tied to the license — HeliBoard is GPL-3.0, the other two are Apache-2.0). Next step after this merges is deciding base+license together and writing the ADR that supersedes [ADR-0001](https://github.com/apexcloudwise/personaspeak/blob/docs/fork-spike-results/docs/adr/0001-thin-ime-over-keyboard-fork.md).

## How it was run

Three `oc-bg -m zai-coding-plan/glm-5.2` workers cloned each candidate at a pinned release commit (resolved via the GitHub releases API, not tag-string sorting) into its own scratch checkout, then vendored `core-personas`/`core-providers` and grafted the strip UI. Claude drove the one AVD (`CityZen_Dev`) serially — install, drive, screenshot, grade — since only one IME can be active at a time. **Every bug the on-device pass found was then fixed by a follow-up GLM worker and independently re-verified on-device by Claude afterward — nothing here is taken on a worker's self-report.** Gboard was restored as default after each pass.

| Candidate | License | Pinned tag | Commit | Published |
|---|---|---|---|---|
| HeliBoard | GPL-3.0 | `v4.0` | `bd48798` | 2026-07-10 |
| AnySoftKeyboard | Apache-2.0 | `1.13-r1` | `8c1db51` | 2026-02-08 |
| FlorisBoard | Apache-2.0 | `v0.5.2` | `2e82060` | 2025-11-28 |

## Comparison

| | HeliBoard | AnySoftKeyboard | FlorisBoard |
|---|---|---|---|
| License | GPL-3.0 | Apache-2.0 | Apache-2.0 |
| Upstream files modified | 4 (+51/-1 after fixes) | 5 (fixes added 0 upstream files) | 4 (2 real), +11/-1 |
| Clean build time | ~7m20s | ~6m30s | ~5m (+ host toolchain fixes) |
| UI toolkit match | Custom Views (native) | Custom Views (native) | Compose (native) |
| Picker works on-device | ✅ (2×2 grid, correct persona names) | ✅ | ✅ (placeholder persona names) |
| Transform works on-device | ✅ (after fix) | ✅ (after fix) | ✅ (after fix) |
| Composing-state replace bug | Fixed (`finishComposingText()` first) | **immune by construction** (doesn't use `selectAll`) | Fixed (`finishComposingText()` first) |
| Builds clean out of the box | ✅ | ✅ | ❌ needs off-repo fixes |
| Suggestion-row competition | real, populated | real, populated | dormant (stubbed in v0.5.2) |

**All three candidates now have a verified, working end-to-end flow** — persona picker, selection, and a clean transform-and-replace — after a second round of commissioned fixes. None of that was taken on faith: every fix was independently reproduced and confirmed on a real emulator afterward.

**Cross-candidate finding #1:** no graft worked cleanly end-to-end on the first on-device attempt, and every worker's initial report over-claimed confidence ("wiring is correct," "unit tests pass") because its tests exercised the pure logic layer (`PromptBuilder`, `FakeProvider`) but never the real `InputConnection` capture/replace path — exactly where each bug lived. Whatever base is chosen, an on-device instrumentation test of that path belongs in the suite.

**Cross-candidate finding #2:** the "leftover text when the last word is still autocorrect-pending" bug turned out to be a genuine pattern, not a one-off — both HeliBoard and FlorisBoard hit it (same root cause, same fix), while AnySoftKeyboard was immune by construction because its replace mechanism doesn't use `selectAll` at all. A real, if secondary, data point on implementation robustness, independent of the base decision.

With all three now functionally equivalent, the deciding factors are what doesn't change with more fixes: **license**, **diff cost/architecture fit**, and **build-maintenance reality**.

---

## HeliBoard

Cheapest-looking diff (4 files, +33 lines), but **the strip was not tappable as first built** —

HeliBoard strip idle

`dumpsys window` showed the IME's touchable region starting at y=1502 while the strip sat at y≈1434–1500. **Fixed:** a commissioned worker refuted its own assigned hypothesis, correctly traced the bug to `LatinIME.onComputeInsets` excluding the strip's height from the touchable-region calculation, and also found a second latent bug — an `<include>` id shadowing the root view's own id, meaning the controller had never actually attached even once touchable. Both fixed in the same file (+51/−1 lines total, still 4 files).

Then a second fix (composing-state replace) landed the same way as FlorisBoard's, below. Verified clean end-to-end afterward:

HeliBoard persona picker working

HeliBoard clean transform result

**Now the most spec-compliant of the three** — the only candidate with the exact persona names (Jeeves, Dr. King Schultz, Sir Humphrey Appleby, Amitabh Bachchan) in a proper 2×2 grid, and a fully verified working transform, at a cost of +51/−1 lines across 4 files.

## AnySoftKeyboard

Comparable diff cost (5 files, +30 lines). The one real integration cost was Java↔Kotlin interop — `:ime:app` had no Kotlin at all, so a small adapter class bridges the vendored `suspend fun rewrite()` into Java. Persona picker worked correctly on-device from the start:

AnySoftKeyboard persona picker

But **tapping transform crashed**: `NoSuchMethodError: No interface method isEmpty()Z in class Ljava/lang/CharSequence`, a build/desugaring interaction (the repo's most recent upstream commit is literally titled "min-api-level23"), not a logic bug. **Fixed** with a one-line, zero-upstream-cost change — `TextUtils.isEmpty(draft.text)` instead of `draft.text.isEmpty()` — avoiding the problematic call entirely rather than chasing the toolchain config. Verified clean afterward:

AnySoftKeyboard fixed transform result

**Bonus finding:** re-tested with a composing/autocorrect-pending draft (the same repro that broke HeliBoard and FlorisBoard) and ASK's replace was clean with no fix needed — its commit path uses `deleteSurroundingText` rather than `selectAll`, and that approach isn't susceptible to the same bug. Not intentional; just a property of its existing implementation that happened to matter.

## FlorisBoard

Cheapest real diff (4 files, +11/-1, only 2 lines of which are the actual code seam) and the cleanest architecture fit — idiomatic Compose, not a bridge:

FlorisBoard strip idle

Persona picker and transform both worked on the first unmodified attempt — the only candidate where that was true — with one bug: **if the last word was still composing/autocorrect-pending when send was tapped, the replace left a leftover fragment.** **Fixed** by calling `ic.finishComposingText()` before `selectAll`/`commitText`, confirmed by reasoning through `InputConnection`'s documented behavior, not just pattern-matching the symptom. Verified clean afterward, retesting the exact composing-state repro:

FlorisBoard fixed composing-state result

Also confirmed from source: FlorisBoard's word-suggestion row really is stubbed at v0.5.2 (`suggest()` returns `emptyList()`) — the persona strip isn't competing with populated suggestions here, unlike the other two.

Against all that: **this checkout doesn't build out of the box.** A pinned `jetpref` snapshot dependency has been purged from Sonatype, and the native module requires a Rust toolchain the host didn't have. Both pre-existing, unrelated to the spike, but they're the "maintainer says it stalled" risk from the checkpoint showing up hands-on. Also, the worker ignored the persona-naming spec and invented placeholder personas ("The Concierge," "The Pirate," etc.) instead of the four specified — not fixed in this pass, a real compliance gap relative to HeliBoard.

---

## Theming and customization

All three grafts correctly skipped theming (out of scope for a graft-cost spike), but "basic" isn't a ceiling for any of them — it's what each strip is currently wired to vs. what it could be:

- **HeliBoard** — reuses `?attr/suggestionWordStyle`, so it automatically matches the user's chosen keyboard theme (HeliBoard has a real `KeyboardTheme.kt` system) — free consistency, but reads as "another suggestion chip" rather than a distinct identity.
- **AnySoftKeyboard** — hardcoded literal colors, completely disconnected from ASK's own theming, which is actually the **most mature of the three** (`KeyboardThemeFactory`, a real community-skin ecosystem). Would clash with any dark/custom theme today; fixing it means wiring into an existing system, not building one.
- **FlorisBoard** — looks the most polished in the screenshots (plain Material3) but isn't wired into FlorisBoard's own theme engine, **Snygg** — a genuinely sophisticated CSS-like stylesheet system (`lib/snygg`, `FlorisImeUiSpec.kt`) that's arguably the highest-ceiling theming system of the three once integrated. Ironically the prettiest by default today, and the most likely to look wrong the moment a user picks a custom skin.

None of the three are actually stuck with "basic" — each has a working theme system in the base that the spike simply didn't plug into. A secondary input to the base decision, not a primary one.

---

## What's in this PR

- `docs/superpowers/specs/2026-07-20-fork-spike-results.md` — the full log (pinned commits, per-candidate detail, all fixes, comparison table, cross-candidate findings).
- `docs/design/mockups/spike/{heliboard,anysoftkeyboard,florisboard}/` — all 34 screenshots across the initial grading and fix-verification passes.

The actual graft code lives outside this repo, in throwaway clones at `~/workspace/scratch/fork-spike/<name>/` on each candidate's own `spike/personaspeak-graft` branch (not pushed anywhere — happy to push them somewhere reviewable if useful, they're currently local-only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive fork evaluation results, including per-candidate workflows, measurable findings, and on-device validation outcomes.
  * Documented follow-up fixes applied per app (input/touch-region behavior, crash from empty text checks, and composing-state replace handling) with re-verification notes.
  * Included emulator grading-pass information, cross-candidate comparisons, theming/customization wiring differences, independent review corrections, and a remaining “Next” checklist item for an outstanding async race.


zaphodis42:
Reviewed by: Gemini 3.1 Pro

## 1. Recommendation: HeliBoard

I recommend **HeliBoard**.

When you fork a keyboard, you are taking on a massive maintenance burden against Android's ever-changing input, inset, and window management APIs. The single most important factor for long-term survival is the health and velocity of the upstream you are merging from. HeliBoard is the only candidate with a vibrant, current upstream (v4.0, Jul 2026). 

While AnySoftKeyboard (ASK) showed better architectural immunity to the composing-state bug, its 74% Java codebase is a significant liability. Building modern, agent-driven Kotlin features (and Compose UI) on top of a legacy Java architecture guarantees chronic friction. 

FlorisBoard must be entirely disqualified; the maintainer has explicitly stated the project is stalled, it fails to build out of the box, and its suggestion engine is just a stub. Forking FlorisBoard isn't getting a head start—it's taking over a dead, unfinished project.

HeliBoard’s initial bugs (the untappable strip, the `selectAll` composing-state issue) were straightforward surface-level fixes that cost only 51 lines of code. That is a trivial one-time integration tax to pay for a modern, actively maintained base.

## 2. Licensing: The GPL-3.0 Paper Tiger

The PR frames HeliBoard's GPL-3.0 license as a severe cost that "ends the README's Apache-2.0 plan outright." I view this framing as unnecessarily pessimistic.

PersonaSpeak is an open-source project showcasing agent-maintained engineering. Releasing the final APK under GPL-3.0 aligns perfectly with this ethos. More importantly, ADR-0001 established strict module boundaries: `core-personas` and `core-providers` are pure Kotlin modules with no Android dependencies. These core modules can—and should—retain their Apache-2.0 licenses as standalone libraries within the repository. The only part of the project that must be GPL-3.0 is the `app` and `keyboard` wrappers. Unless the unstated goal is to eventually close-source the keyboard or sell it as a proprietary enterprise SDK, adopting GPL-3.0 for the host application is an acceptable and entirely standard trade-off for utilizing the best available open-source base.

## 3. Future Maintainability 

The spike results correctly highlight that FlorisBoard's architecture (Compose) is the cleanest fit, but severely underweights the danger of its "stalled" status. A keyboard with stubbed predictions and unbuildable dependencies is not a foundation. Furthermore, when FlorisBoard eventually implements word suggestions, they will immediately clash with the persona strip for vertical space—a conflict the spike couldn't measure because the feature doesn't exist yet.

AnySoftKeyboard offers a mature prediction engine, but the ongoing upstream merge burden will be high. As ASK continues to evolve its Java codebase, bridging the gap to PersonaSpeak's pure Kotlin architecture will require constant, manual translation and fighting legacy Android paradigms.

HeliBoard's maintainability risk is its 31% C++ codebase (the AOSP dictionary). However, PersonaSpeak's core value—the AI rewrite engine—does not need to interact with the traditional n-gram dictionary. The AI logic lives in the Kotlin layer, safely separated from the C++ core, meaning you rarely (if ever) need to touch the native code.

## 4. Weaknesses in the PR's Reasoning

I explicitly disagree with the PR's reasoning in two areas:

*   **Equating Diff Size with Architectural Fit:** The PR leans heavily on line-count diffs (+11 for FlorisBoard vs +51 for HeliBoard) to evaluate the "graft cost." In a codebase of over 100,000 lines, a 40-line difference is statistical noise. HeliBoard's fixes were isolated to window insets and input connection edge cases—standard Android UI plumbing. Measuring invasiveness by line count rather than the semantic complexity of the seam is a flawed metric.
*   **Under-evaluating the Composing-State Bug:** The PR treats the composing-state bug (where `selectAll` + `commitText` leaves a leftover fragment) as a "secondary data point." This is a fundamental flaw in how HeliBoard and FlorisBoard interact with Android's `InputConnection`. ASK avoided it completely because it uses `deleteSurroundingText`. While the spike proved the bug is patchable via `finishComposingText()`, the PR fails to recognize that ASK's underlying text manipulation architecture is inherently safer for a project whose entire purpose is text replacement. The PR should have weighted ASK's architectural robustness much higher than FlorisBoard's trivial +11 line diff.

Despite ASK's superiority in text replacement mechanics, HeliBoard's modern stack and active upstream make it the only sustainable choice for a small team over a multi-year horizon.

zaphodis42:
Reviewed by: OpenAI Codex (GPT-5)

# Fork-base decision memo

## Top-line recommendation

**Pick AnySoftKeyboard at the evaluated `1.13-r1` / `8c1db51` line.** It is the best risk-adjusted base for a daily-driver PersonaSpeak keyboard: mature autocorrection and next-word prediction, built-in gesture typing, the strongest visible automated-test and release machinery of the three, an upstream release from February 2026, and Apache-2.0.

The Java-to-Kotlin boundary is real but narrow and controllable; the spike keeps it in a new PersonaSpeak package plus one Kotlin adapter (`ime/app/src/main/{java,kotlin}/com/anysoftkeyboard/personaspeak/` at `7edc46b`).

HeliBoard is the runner-up. It is the freshest and visibly active upstream, but its GPL cost is not buying a decisive product advantage over ASK, and its advertised glide typing depends on a separately obtained closed-source library (`~/workspace/scratch/fork-spike/heliboard/README.md`, "Glide typing"). FlorisBoard is not a credible daily-driver base yet: its current upstream README still says it is beta and explicitly says word suggestions and spell checking are absent, while the pinned stable checkout contains the same stub (`LatinLanguageProvider.suggest() = emptyList()` at `2e82060`).

Confidence is **medium**, not high. The spike proves that all three can host a demo strip. It does not yet prove safe asynchronous replacement, production-provider behavior, daily-driver typing quality, or actual upstream-update cost.

## 1. Base choice: AnySoftKeyboard

The deciding evidence is broader than the graft diff:

- ASK already supplies the product features PersonaSpeak promises during onboarding: word and next-word suggestions, configurable autocorrection, and gesture typing. This is stated in the upstream `README.md` at `8c1db51` and the current ASK repository README. That matters more than whether the PersonaSpeak row takes two or twenty lines to mount.

- ASK has materially stronger visible regression infrastructure. At the pinned checkout, `.github/workflows/checks.yml` shards unit tests, coverage, lint, assembly, and a deployment dry run; a filename-based inventory finds roughly 240 test sources. The comparable pinned HeliBoard checkout exposes 43 test sources and one unit-test workflow (`.github/workflows/build-test-auto.yml`); FlorisBoard exposes 19 and its Android workflow only runs `clean assembleDebug`. Counts are crude, but the workflow difference is not.

- ASK documents continuous alpha/beta promotion and staged stable rollout, and its latest stable release is `v1.13-r1` from 8 February 2026. That is current enough to reject the idea that HeliBoard is the only maintained choice.

- The spike's ASK integration does not modify an existing runtime Java class. Its five "upstream files" are Gradle/catalog wiring, one layout include, and `.gitignore`; the bridge and view are additive (`0bc4cff..7edc46b`). In production, `.gitignore` is not an integration seam at all. By contrast, HeliBoard modifies the central `LatinIME.java`, including `onComputeInsets`, and FlorisBoard inserts directly into `TextInputLayout.kt` (`dae4556..6b24ed3`).

- ASK's Java/Kotlin interop cost is overstated as a strategic liability. `PersonaRewriteBridge.kt` is 84 lines and converts one suspend API to callbacks. PersonaSpeak-specific code remains isolated, satisfying `AGENTS.md`'s enduring rule that inherited code should not be peppered with product logic.

I therefore disagree with the results log's implication that ASK has the "costliest interop" in a way that should weigh heavily against it (`docs/superpowers/specs/2026-07-20-fork-spike-results.md`, final comparison). It has the most obvious language boundary, but arguably the least invasive runtime seam: no inherited input-service class was edited.

I also disagree with treating HeliBoard's correct persona names and FlorisBoard's placeholders as base-selection evidence. The results log calls HeliBoard "the most spec-compliant" and repeatedly calls FlorisBoard's invented names a "real compliance gap." Those names live in wholly new spike-owned files: `PersonaPresets.kt` and `PersonaStrip.kt`. They measure worker instruction-following, not host architecture. ASK is actually the only graft that loads the four personas from YAML assets; HeliBoard and FlorisBoard construct them in code.

### Why not HeliBoard?

HeliBoard is a respectable alternative and may win a subjective typing-quality bake-off. It is current, actively developed, builds successfully at the pinned release, and has a competent existing prediction system. The graft also demonstrated that PersonaSpeak can integrate with its theme tokens.

However:

- Its glide typing requires a closed-source external library rather than being present in a clean FOSS build. That conflicts with the onboarding design's unqualified reassurance that the replacement keyboard includes glide typing (`docs/superpowers/specs/2026-07-20-keyboard-ux-design.md`, §4).

- Its PersonaSpeak seam reaches into `LatinIME.java` and changes touchable-inset calculation. The first on-device pass found that the strip sat outside the IME's touchable region, and the fix added host-specific behavior to `onComputeInsets` (`e811e35`). That is precisely the sort of central inherited plumbing likely to require attention during upstream merges and unusual IME modes.

- It brings GPL-3.0 constraints without evidence that ASK's typing experience is inadequate.

### Why not FlorisBoard?

FlorisBoard is the most attractive codebase for writing PersonaSpeak UI. It is Kotlin/Compose-native, the strip mounts through a two-line seam, and Snygg offers a strong theming ceiling. Those are real advantages.

They do not compensate for the missing keyboard fundamentals. The pinned source verifies that word suggestions are stubbed (`app/src/main/kotlin/dev/patrickgold/florisboard/ime/nlp/latin/LatinLanguageProvider.kt` at `2e82060`), and the current FlorisBoard README still says word suggestions and spell checking are not included. PersonaSpeak has committed to replacing the user's daily keyboard. Choosing FlorisBoard would make PersonaSpeak responsible for completing and maintaining the hardest part of the keyboard rather than inheriting it.

The clean Compose seam is therefore a local optimum attached to the wrong product baseline.

## 2. Licensing: GPL-3.0 is acceptable, but not worth paying here

GPL-3.0 is **not inherently disqualifying** for PersonaSpeak. This is already intended to be FOSS, the privacy story benefits from inspectable keyboard code, and GPL does not prevent selling the app or operating a separate hosted service. Apache-2.0 code can be combined into a GPLv3 program (ASF describes this compatibility as one-way).

The PR's shorthand that HeliBoard "relicenses the whole app" is directionally useful but imprecise. Distribution of a combined HeliBoard-based Android application would need to satisfy GPLv3 for the combination and provide complete corresponding source. The FSF treats static or dynamic linking as a combined work covered by GPL terms. This memo is engineering advice, not a substitute for counsel before release.

The practical cost is option value:

- Every shipped Android client component and future SDK must be GPL-compatible.
- Binary releases need disciplined corresponding-source, build-script, notice, and installation-information handling.
- A future closed-source client tier, proprietary embedded model, or incompatible analytics/crash SDK becomes difficult or impossible to ship in the same app.
- The team must audit HeliBoard's GPL, Apache, icon, dictionary, and optional glide-library licensing as an actual distribution set, not just copy one `LICENSE` file.
- Copyleft would also make future base migration more awkward if PersonaSpeak-specific Android client code becomes tightly combined with GPL-covered HeliBoard code. `core-personas` and `core-providers` can remain logically isolated, but distribution boundaries and notices must be designed deliberately.

Those are manageable costs if HeliBoard is uniquely superior. The evidence does not show that it is. ASK retains Apache-2.0 while offering mature typing, gesture input, deeper test/release automation, and a similarly small graft. I would not spend the GPL option value merely for HeliBoard's newer release date and somewhat more modern host.

One correction to the framing: the checked-out `README.md:69-75` does not currently *promise* Apache-2.0. It says "license TBD — leaning Apache-2.0" and explicitly notes that the fork base may make it GPL-3.0. The licensing decision is still open in repository prose.

My conclusion:

- GPL-3.0 is not a veto.
- It is a meaningful, permanent product constraint.
- HeliBoard has not demonstrated enough superiority over ASK to justify accepting that constraint.
- If a controlled daily-driver bake-off later shows that ASK's typing quality is materially worse, HeliBoard should be reconsidered and GPL accepted deliberately. The present spike does not show that.

## 3. Future maintainability

### Upstream health

**AnySoftKeyboard:** The strongest long-term maintenance bet. The current GitHub repository reports 8,870 commits, a February 2026 stable release, and automated alpha/beta/stable promotion. The pinned branch's `.github/workflows/checks.yml` and large test inventory support that operational maturity. Its old Java core is not fashionable, but stable keyboard behavior and tests are more valuable than fashionable code in the component whose failure mode is "the phone cannot type."

**HeliBoard:** Clearly healthy and the most current. Release `v4.0` at `bd48798` was published 10 July 2026, with multiple contributors across the preceding alpha/beta train. It remains a good fallback if ASK fails a production typing bake-off.

However, the results log's claim that HeliBoard has "the deepest commit history" is contradicted by its own checkpoint table and current GitHub data. ASK reports 8,870 commits, while the checkpoint attributes about 2,480 to HeliBoard (`docs/superpowers/specs/2026-07-20-fork-spike-checkpoint.md`, Candidate evidence). "Healthiest and most current" is defensible; "deepest commit history" is not.

**FlorisBoard:** The PR overstates one part of the case and understates another. The 2023 maintainer post really did say development had "grinded to a halt" and that maintaining it alone was no longer possible (discussion #2314). I personally verified that primary post.

But that is historical evidence, not proof the project is still halted. `v0.5.2` shipped in November 2025, the repository shows 2026 activity, and the project now has multiple contributors. The PR should not use a 2023 statement as if nothing happened afterward.

The stronger present-day rejection is product readiness. The current upstream README still labels FlorisBoard beta and says word suggestions and spell checking are not in current releases. PersonaSpeak has already committed to a daily-driver keyboard with autocorrect. Inheriting a keyboard that does not yet provide it means owning the hardest keyboard subsystem.

The Floris build discussion also needs correction. The purged `jetpref` snapshot is a genuine reproducibility defect, directly verified by the one-line version change in `gradle/libs.versions.toml` at `dae4556`. A missing Rust installation is not evidence of neglect: Rust is an explicit prerequisite in the pinned checkout's `CONTRIBUTING.md:48-67`. The results log combines these as "off-repo intervention" and treats both as confirmation of under-maintenance; only the dead dependency supports that conclusion.

### Upstream merge burden and graft durability

The spike's headline rule — "upstream lines modified are rent paid forever" — is useful but too weak to decide long-term cost by itself. Merge-conflict probability depends on *where and how often upstream edits occur*, not raw line count.

**HeliBoard:** four-file, +51/-1 footprint includes 39 changed lines in `LatinIME.java`, a central service class, and custom `onComputeInsets` behavior (`bd48798..cfe65b1`). That is a hotter, more semantically coupled seam than ASK's layout/build wiring even though ASK's table says "5 files." The inset change also needs testing in floating, split, hardware-keyboard, fullscreen/extract, edge-to-edge, rotation, and popup modes — the spike exercised one portrait emulator path only.

**AnySoftKeyboard:** its five-file/+30 count includes `.gitignore` and 12 version-catalog lines; neither should be weighted like a runtime-service modification. Its production seam is essentially a layout include plus Gradle dependencies. The one observed crash was in new PersonaSpeak code (a JDK-default `CharSequence.isEmpty()` call unavailable on the device, fixed with `TextUtils.isEmpty()` at `7edc46b`) — cheap, and not evidence that ASK's host architecture is hostile to the graft.

**FlorisBoard:** has the cleanest immediate UI seam (one import, one `PersonaStrip()` call). Any typical upstream restructure of the keyboard's main Compose hierarchy could conflict, but the conflict would be obvious and small. Nevertheless, a cheap seam onto an unfinished host is still an expensive product — the likely long-term burden is completing prediction, autocorrection, spelling, and possibly glide typing while simultaneously tracking upstream's own work in those areas.

New spike-added files are not free to maintain even when cheap for conflict resolution: roughly 400 lines of host-specific UI/adapter code for HeliBoard and ASK, 340 for FlorisBoard, all needing lifecycle, privacy, accessibility, localization, error, and editor-compatibility ownership forever.

**The missing merge experiment:** none of the three branches was rebased onto a later upstream commit; no file-churn analysis or trial upgrade was performed. The PR can claim one-time graft size — it cannot yet claim measured ongoing merge cost. Before the ADR, run one synthetic upstream-update exercise for ASK and HeliBoard: replay each graft from its pinned stable tag onto current upstream head, record textual conflicts separately from behavioral/build fixes, rebuild, and record time-to-green.

## 4. Where the PR's reasoning is weak or wrong

### The more serious cross-candidate gap: asynchronous stale-field replacement

**All three grafts can apply an asynchronous provider result to text that changed after the request began, and potentially to a different current field if focus changes while the provider is running. This can overwrite newer or unrelated user text.** This is more serious than the composing-span defect the PR fixed, because a real network provider creates a much larger delay than `FakeProvider`, making the race ordinary rather than theoretical.

- **HeliBoard** (`.../PersonaSpeakController.kt` at `cfe65b1`): captures a draft from an `InputConnection`, awaits the provider, then unconditionally calls `finishComposingText()`, `selectAll`, `commitText()` on the *retained* old connection — with no check that the same editor is still active, the field contents still match, the user hasn't typed more, selection hasn't changed, or the connection is still valid.

- **AnySoftKeyboard** (`.../PersonaStripView.java` at `7edc46b`): captures a draft, awaits the provider, then *re-resolves* the current connection and calls `deleteSurroundingText(draft.beforeLength, draft.afterLength)` using lengths captured from the old draft, without checking editor identity or current content. If focus moved to another field, it can delete text from that new field using stale lengths. The PR praises ASK's `deleteSurroundingText` approach as "immune by construction" to the composing bug — it may avoid that one reproduced case, but its stale-length behavior creates a more serious race than the bug it fixed. (There's also a weaker selected-text inconsistency: `captureDraft()` returns `selectionStart/End = -1`, but `commitResult()`'s explicit-selection branch requires non-negative positions, so it falls through to `deleteSurroundingText(0, 0)` relying on the editor's selection still being active.)

- **FlorisBoard** (`.../PersonaStrip.kt` at `6b24ed3`): reads `editorInstance.activeContent.text`, awaits the provider, then unconditionally finishes composing, selects all, and commits — resolving the current input connection only at completion time. If focus changed, this replaces all text in the new field; if the user kept typing in the same field, it replaces that newer content.

This directly conflicts with the UX design's own rules in `docs/superpowers/specs/2026-07-20-keyboard-ux-design.md` §3 — "the draft stays visibly intact," "ask before replacement" is the default, "the user's own words are never destroyed without a tap." The spike bypasses the result card and commits immediately, so the only tap happened before an asynchronous interval during which the target text could change. Production needs, at minimum: a request token tied to active editor/field identity, cancellation on focus/field change or teardown, a captured-draft equality check before replacement, a result preview that resolves and validates the target at tap time, and explicit handling when an editor can't expose enough identity to validate safely.

### "Functionally equivalent" is too strong

The three candidates are equivalent only in a narrow emulator demo using `FakeProvider` and immediate replacement. Actual text-capture and replacement semantics differ (selection preference, capture-length fallback, deleteSurroundingText vs. selectAll), and none implements the designed result card, ask-before-replace default, real provider/network path, key storage, cancellation UI, or full error states. The PR's "immune by construction" language for ASK should be softened to "not susceptible in the tested editor and path."

### "All three now have a verified end-to-end flow" needs qualification

I personally verified the final graft code, git histories/commit IDs, diffstats, build logs/APK hashes in `SPIKE_REPORT.md`/`EVIDENCE.md`, and the three post-fix screenshots. I did **not** independently replay the emulator sessions or cryptographically tie each screenshot to the stated APK commit. The scratch reports say post-fix device verification was pending/unavailable at spike time; the PersonaSpeak results log says that verification was later performed — so the final on-device claims rest on the results log and screenshots, not on independent re-verification by me.

### The spike's own metric was applied inconsistently

ASK's "five upstream files" includes `.gitignore` (pure hygiene); FlorisBoard's +11/-1 includes an unrelated `jetpref` dependency repair counted alongside two "real" lines; HeliBoard commits `_vendor_src/` reference material while the others don't. This doesn't invalidate the spike, but the comparison table should be read as an approximate conflict-surface inventory, not a normalized maintenance-cost benchmark. By a fairer accounting (separating required runtime seam from build wiring, hygiene, unrelated repair, and new host-specific code), ASK likely has the lowest-risk inherited runtime seam even though it shows the largest raw file count.

### The PR overweights first-pass worker mistakes

Not all first-pass bugs are evidence about the host: FlorisBoard's placeholder persona names and ASK's `CharSequence.isEmpty()` crash are worker/toolchain mistakes in new spike-owned code. HeliBoard's touchable-region problem and the HeliBoard/FlorisBoard composing-text problem are genuine host/`InputConnection` integration-contract issues. The PR sometimes lets bug count and severity stand in for host quality; a better reading separates spike-implementation quality from integration-contract risk that production must test regardless of base.

### Other things the spike evidence didn't cover but should have

- **Daily-driver typing quality was never actually compared** (autocorrect accuracy/latency, next-word prediction, personal dictionary, multilingual switching, gesture-typing quality, one-handed/split/floating modes, battery/ANR behavior). The recommendation trades HeliBoard's fresher UI for ASK's mature typing engine on the strength of feature lists, not measured use.
- **Privacy/network trust boundary was not exercised at all** — disabling transformations in password/PIN fields, incognito/no-personalized-learning respect, keeping provider text out of learning dictionaries and crash logs, cancellation on navigating to a sensitive field, clipboard/history retention of generated text. This deserves an ADR-level decision regardless of base, and reinforces keeping `core-providers` narrowly isolated per `AGENTS.md`.
- **Theming cost is asserted, not measured** — "roughly comparable effort" for ASK vs. FlorisBoard theming is an inference, not something verified by actually implementing a themed variant in each.
- **Build-health reasoning conflates two different things** — FlorisBoard's purged `jetpref` snapshot is real evidence of neglect; a missing Rust toolchain (explicitly documented as a prerequisite) is normal setup, not a second maintenance defect, and shouldn't be double-counted.

### Required follow-up before the ADR is final

I would not reopen the decision to fork a full keyboard. I would select ASK as the presumptive base and require: (1) replay of the ASK graft onto current upstream head to measure real merge cost, with HeliBoard replayed as fallback comparison; (2) a short multi-day typing bake-off (autocorrect, gesture typing, multilingual, stability); (3) an editor-token/draft-validation guard implemented and tested against focus/text changes during an in-flight request; (4) exercise of password, incognito, WebView, Compose-host, and long-draft editor cases; (5) one dark/custom theme and one landscape/small-screen layout verified; (6) a real Apache/GPL dependency-and-distribution audit before finalizing the licensing consequence.

The spike did useful work — it disproved "a small compile-green graft means it works on-device" and surfaced a real composing-state failure pattern. It should not be stretched into proof that UI line count is the dominant lifetime cost. For this project, the inherited keyboard's typing completeness, regression machinery, licensing headroom, and safe text-replacement architecture dominate. On the evidence available, that points to **AnySoftKeyboard**.

zaphodis42:
Reviewed by: Claude Sonnet 5 (this session — ran the spike, all fixes, and both independent reviews above)

## Closing note, after Gemini and Codex

I asked for two independent reviews precisely because I didn't want this PR's own framing to be the last word. They landed on opposite recommendations — Gemini: HeliBoard, Codex: AnySoftKeyboard — and both pushed back on this document's reasoning in ways I checked against the actual code rather than took on faith. Two of Codex's claims are now verified and corrected in the results doc: the README doesn't promise Apache-2.0 (it's "TBD — leaning," explicitly conditional on this decision), and HeliBoard doesn't have the deepest commit history (ASK's 8,870 vs. ~2,480, per the checkpoint's own table). Codex's async stale-field race finding is real too — I read all three transform pipelines myself and confirmed it: every candidate captures the draft (and, for two of them, the `InputConnection`) before awaiting the provider, then commits unconditionally afterward. `FakeProvider`'s fixed 400ms delay hid this in every test I ran; a real provider's latency would not.

**Given all of that, my own recommendation is AnySoftKeyboard** — I land closer to Codex than to Gemini, for one specific reason neither the original PR nor Gemini weighed properly, which I checked myself before writing this:

**HeliBoard's glide typing requires a closed-source library not included in the FOSS build** (its own README, verbatim: "Glide typing (*only with closed source library* ☹️) — library not included in the app, as there is no compatible open source library available"). AnySoftKeyboard lists gesture typing as a plain native feature, no such caveat. This matters specifically for PersonaSpeak: the onboarding design this spike is downstream of makes an explicit, unqualified reassurance triple — "Autocorrect · Glide typing · Works offline" — to answer the exact objection a full-keyboard-replacement pitch has to overcome. Forking HeliBoard means either quietly breaking that promise (no glide typing in the FOSS build) or bundling a closed-source blob into a project whose entire pitch includes "inspect our code, we're not a diary." ASK delivers the same promise natively, Apache-2.0, no compromise. I don't think either the original spike or Gemini's review weighed this heavily enough, and it's a bigger deal to me than a few dozen lines of diff either way.

On the diff-size argument specifically: I think Codex's framing is more internally consistent than Gemini's. Gemini says diff-line-count is "statistical noise" in a 100k-line codebase, then still leans on HeliBoard's small headline diff as reassurance. Codex's point — judge the seam by *what kind of file* it touches, not by line count in either direction — is the sharper tool. HeliBoard's fix touches `LatinIME.onComputeInsets`, a central, behaviorally significant service method; ASK's five files are almost entirely Gradle/catalog/`.gitignore` hygiene around net-new code. By that read, ASK has the least invasive seam of the three despite the higher file count, which is the opposite of how this PR originally framed it — and I think that original framing was too quick to equate "small diff" with "safe integration."

On licensing, I agree with both reviewers that GPL-3.0 isn't an automatic veto — but I lean toward Codex's more cautious read over Gemini's "paper tiger" framing. `core-personas`/`core-providers` staying pure and portable was never in question; the actual constraint lands on `app`/`keyboard` at distribution time, and "license TBD" in the README means there's no sunk cost being protected by staying Apache-2.0 — but there's also no reason to spend that option value on a base that isn't clearly better on the merits that matter most (typing completeness, without a closed-source dependency).

On FlorisBoard: all three of us agree it's out for now, for the same reason — cleanest architecture and the highest theming ceiling don't matter if the base doesn't type well yet, and it doesn't build without off-repo intervention today. That one isn't close.

**What I'd actually do next**, combining both reviewers' follow-up suggestions with my own read: pick AnySoftKeyboard as the working base, but treat it as provisional pending (1) a short real-usage typing bake-off — autocorrect/prediction/gesture-typing quality, not just feature lists, since that's the one claim in this whole thread nobody has actually measured hands-on; and (2) a real fix design for the async stale-field race before any real provider replaces `FakeProvider`, independent of which base wins. Both of those matter more to me than anything else raised in either review.

PR #8, opened 2026-07-20
ci: enforce the PATCHNOTES gate on every PR

zaphodis42:
## What

A new `patchnote-gate` job in `.github/workflows/ci.yml`, triggered on `pull_request`, that fails the build when the PR's diff doesn't touch `PATCHNOTES.md`. Escape hatch: a `no-patchnote` label skips the check with a loud `::notice`. Scope is deliberately "touched, not good" — quality stays with the reviewer.

This PR also dogfoods the gate: it adds the matching entry to `PATCHNOTES.md` itself, because otherwise it would fail its own check on arrival.

## Why

Closes #4. The Definition of Done requires every merged PR to add a `PATCHNOTES.md` line, but until now nothing enforced it — PR #3 shipped without one until review caught it. Client-side hooks (husky / pre-commit) are the wrong tool: bypassable with `--no-verify`, absent from CI and fresh agent checkouts, invisible to reviewers. A CI gate can't be skipped by accident.

The `no-patchnote` escape hatch exists for genuine exceptions (pure reverts, meta-fixes on the gate itself). The skip is announced, never silent — per the house rule on loud skips. Not architectural, so no ADR; the gate's posture follows the CI-gate lineage inherited from darkmill ADR-0031.

## Evidence

- **Commands run + output:**

  YAML validity — `actionlint` is not installed locally, so the task-spec fallback (`python3 -c "import yaml, ..."`) was used:

  ```
  $ python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/ci.yml')); print('YAML OK; jobs:', list(d['jobs'].keys()))"
  YAML OK; jobs: ['personas-and-cli', 'patchnote-gate']
  ```

  Shell-logic proof — the core check was extracted to a local script (`/tmp/patchnote_check.sh`, mirrors the workflow step byte-for-byte) and run against three sandbox scenarios. The script's `git diff --name-only <base>...HEAD | grep -qx 'PATCHNOTES.md'` is identical to what lands in CI; the label branch mirrors `contains(github.event.pull_request.labels.*.name, 'no-patchnote')`:

  ```
  === Scenario 1: feature branch TOUCHES PATCHNOTES.md  (expect: PASS, exit 0) ===
  --- git diff --name-only main...HEAD (sanity):
  PATCHNOTES.md
  --- running check (cwd=sandbox):
  NOTICE: PATCHNOTES.md is in the diff - gate satisfied.
  exit=0

[...]

Issue #7, opened 2026-07-20
Decide the fork base + license, and write the ADR superseding ADR-0001


zaphodis42:
## Why

The fork spike (docs/superpowers/specs/2026-07-20-fork-spike-results.md) produced the evidence; it deliberately does not pick the base. That's the owner's call, and it needs a formal ADR that supersedes ADR-0001. Two independent reviews reached opposite recommendations (Gemini: HeliBoard; Codex: AnySoftKeyboard), so this is a genuine decision, not a rubber stamp.

## Decision sequence (recommended)

1. **Lock the license first** — it's the only irreversible axis. HeliBoard is GPL-3.0 (makes the whole distributed app GPL, one-way door); ASK and FlorisBoard are Apache-2.0. The README currently says "license TBD — leaning Apache-2.0". If Apache matters at all, HeliBoard is out regardless of how good its picker is. This is a business/legal call — get real legal review before committing to GPL.
2. **Run the typing bake-off** (#5) between the surviving candidates — keyboard feel is the product and it's the one unmeasured axis.
3. **Design the stale-field race fix** (#6) — base-independent, must precede any real provider.
4. **Write the ADR** — record base + license together, superseding ADR-0001. A decision that lives only in a chat transcript does not exist.

## Provisional lean

AnySoftKeyboard: matches the owner's in-thread lean, keeps Apache, builds clean out of the box, deepest commit history, most mature prediction engine; its worst trait (Java<->Kotlin interop) is a one-time tax already paid in the spike. But do not finalize the ADR before the bake-off (#5) and the race-fix design (#6).

## Blocked by

- #5 (typing bake-off)
- #6 (stale-field race fix design)

Issue #5, opened 2026-07-20
ASK typing sanity check (descoped from the 3-way bake-off)


zaphodis42:
## Descoped 2026-07-21

The fork base + license decision (ADR-0003) is **AnySoftKeyboard under Apache-2.0**. That retires the original 3-way comparative bake-off this issue described:

- **License removes HeliBoard** (GPL-3.0) from contention outright.
- Among the two Apache options, **ASK already wins the typing axis from facts in hand**: FlorisBoard's suggestion engine is stubbed at v0.5.2 (`LatinLanguageProvider.suggest()` returns `emptyList()`), while ASK has the most mature prediction engine of the three. FlorisBoard also doesn't build out of the box. A full 500-phrase × 3-keyboard run would spend most of a day confirming that a real engine beats a stubbed one.

## What remains (the part still worth doing)

A **provisional, single-keyboard sanity check**: does ASK's inherited typing engine clear the bar — not a downgrade users would notice vs a competent stock keyboard? This is a **fast-follow, not a blocker on #7** — adopting ASK is course-correctable (the fork is survivable; `core-*` transplant unchanged, per ADR-0001's escape hatch).

Scope:
- ASK only. No cross-keyboard tap-mode-per-render.
- A **trimmed corpus** (~20–30 phrases: the academic set sampled + the domain messages), not the full 500.
- **Fix the `clear_field()` bottleneck first** regardless — 256 `KEYCODE_DEL` adb round-trips per phrase is a real scaling problem (select-all + single delete instead).
- Metrics that matter for "good enough": autocorrect false-correction rate, gesture first-pass accuracy, prediction acceptance.

## Salvage from PR #9

PR #9 (the 3-way harness, closed) has reusable pieces: the corpus, the metrics, the tap-mode approach, and the `clear_field` fix. Pull those forward into the ASK-only check rather than rebuilding.

Estimate: a couple hours, deferrable. Not on the critical path.

zaphodis42:
Status update, not a resolution — this issue should stay open and the base-and-license ADR should still not finalize.

**PR #9** builds the harness (corpus, adb replay, Soukoreff-MacKenzie metrics, results template) and, as of the latest commits, has been smoke-tested for real:

- Installed stock F-Droid builds of all three candidates on the emulator (not the graft-spike debug builds that happened to already be sitting there — those would have measured our own graft, not the base).
[...]

Issue #4, opened 2026-07-20
Enforce the PATCHNOTES gate in CI


zaphodis42:
## Why

The Definition of Done requires every PR to add a line to `PATCHNOTES.md`, but nothing enforces it — PR #3 shipped without one until review caught it. Client-side hooks (husky/pre-commit) are the wrong tool: bypassable with `--no-verify`, not installed in CI or fresh agent checkouts, invisible to reviewers.

## What

Add a job to `.github/workflows/ci.yml`, on `pull_request`, that fails if the PR diff doesn't touch `PATCHNOTES.md`.

- Compare against the PR base: `git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -qx PATCHNOTES.md`.
- Escape hatch: skip the check if the PR carries a `no-patchnote` label (pure reverts, etc.) — a loud, visible skip, per the house "loud skips, never silent ones" rule.

## Scope

- Verifies the file was *touched*, not that the line is good/in-voice — that last mile stays with the non-author reviewer. Catching "forgot entirely" is the point.
- Separate concern from any content PR; lands on its own.

Follows the CI-gate posture inherited from darkmill ADR-0031.

PR #10, opened 2026-07-20
docs: spec the stale-field race guard for the transform pipeline

### LINKED ISSUE #6: Design a fix for the async stale-field race in the transform pipeline
## Why

Independent review (Codex, verified by the owner against the code) found all three candidate transform pipelines share an unguarded async race: each captures the draft — and, for HeliBoard/FlorisBoard, the InputConnection — before awaiting the provider, then commits the result unconditionally afterward. No check that the same field is still focused, contents unchanged, or connection still valid.

FakeProvider's fixed 400ms delay hid this in every on-device test. A real network provider's multi-second, variable latency makes it ordinary, not an edge case. This directly violates the UX design's "the user's own words are never destroyed without a tap" — the only tap happens before the unguarded window.

AnySoftKeyboard's variant is arguably sharper: it re-resolves the connection after the await but replaces via deleteSurroundingText(beforeLength, afterLength) using lengths captured from the old draft — so a focus change can delete from a different field using stale lengths. (This is why the doc's "immune by construction" was softened to "not susceptible on the tested path".)

## What's needed

A design (then implementation, once a base is chosen) for an editor-identity / generation token that invalidates a pending rewrite if the target field or its contents change before the result lands. Sketch:

- Capture an identity token at request time (editor field id + a content/selection hash + a monotonic generation counter).
- On result, re-validate: same editor focused, same generation, content still matches captured draft, connection valid. If any fails -> discard the result (surface a non-destructive "couldn't apply" state per the UX design), never commit.
- Decide the cancellation UX (in-flight rewrite superseded by new typing / focus change).

## Base-independent

Production requirement regardless of which keyboard wins; must be designed before any real provider replaces FakeProvider. Belongs in the base ADR's "before we ship" list.

Ref: results doc "Independent review" section, docs/superpowers/specs/2026-07-20-fork-spike-results.md.

zaphodis42:
## What

Design spec for **issue #6** — an editor-identity / generation-token guard at the keyboard↔`core-providers` seam that invalidates a pending rewrite when the target field or its contents change before the result lands. The race was found by Codex's independent review of the fork spike (verified against the code), is present in all three candidate pipelines, and is hidden today only by `FakeProvider`'s fixed 400ms delay.

New file: \`docs/superpowers/specs/2026-07-21-stale-field-race-design.md\`. **Code-free.** No \`keyboard/\`, \`app/\`, or \`core-*/\` touched; persona schema unmoved.

Closes #6.

## Why

The [keyboard UX design](../tree/main/docs/superpowers/specs/2026-07-20-keyboard-ux-design.md) lays down one rule the whole result-card hangs from: *the user's own words are never destroyed without a tap.* The only tap (the transform button) happens **before** the unguarded async window — there is no re-validation when the provider result lands. Under \`FakeProvider\` the rule holds by coincidence; under a real network provider's multi-second latency it does not hold at all. This spec restores it by making the commit conditional on the editor state still matching what was captured, and discarding (non-destructively) when it doesn't.

This is a **spec, not an ADR.** The binding base+license decision rides with issue #7's ADR, which this doc informs. The contract is base-independent: pure Kotlin, no Android imports — whichever base #7 picks only supplies the \`InputConnection\` adapter.

## Evidence

- **Commands run + output:** \`rg -i 'leverage|utilize|seamless|empower|robust|best-in-class|please note that'\` over the new doc and PATCHNOTES → no matches (VOICE banned-words clean). \`git diff --stat\` → \`docs/superpowers/specs/2026-07-21-stale-field-race-design.md\` (new, 445 lines) + \`PATCHNOTES.md\` (+10). No other paths touched.
- **Screenshots / goldens:** N/A — design doc, no UI, no prompts.
- **Journey recording:** N/A — no flow change.
- **Graded by:** **PENDING — non-author review required.** The reviewer is Claude (different model family than the author, GLM-5.2). **This PR must NOT be merged until that review lands as a PR comment.** Per the Definition of Done in AGENTS.md, the author does not produce the verdict that its own work is correct.

## Patch note

Added at the top of \`PATCHNOTES.md\`:

> ## 2026-07-21 — The gate that watches the gate's more dangerous cousin (PR #10)
>
> - Speced the editor-identity guard that stops a slow provider from rewriting the wrong field — or the right field with somebody else's freshly-typed words. The race was always there; \`FakeProvider\`'s brisk 400ms just kept it politely off-stage. No code lands here, and no base is picked — the fork-base ADR still owns that — but the contract is now written down where the implementation ticket can't misplace it. \`FakeProvider\`, for its part, remains a committed thespian and an unreliable witness.

## Checklist

- [x] Tests ride with the code (regression test, if this fixes a bug) — **N/A: design doc only, no product code.** No tests to ride; the on-device instrumentation test plan that the fork-spike results doc calls for is specced in §5 and is the responsibility of the implementation ticket (one per base, after #7 lands).
- [x] Goldens updated and explained (never silently changed) — **N/A: no prompts, no schema, no UI.** Goldens are byte-identical because nothing that feeds them moved.
- [x] Docs in this PR (ADR if architectural; schema consumers if schema moved) — **this PR *is* the doc.** Deliberately not an ADR: the binding base+license decision rides with issue #7's ADR, which this spec informs.
- [x] \`desktop/personaspeak.py --list\` + golden tests pass, if \`personas/\` or prompt code was touched — **N/A: neither touched.** \`personas/*.yaml\` unchanged; \`schema_version\` stays 1.
- [x] Patch-note entry committed to \`PATCHNOTES.md\` (not just pasted above) — yes, in the file at the top, PR #10 heading.
- [x] Graded by a non-author (or "review skipped: <reason>" stated above) — **PENDING.** Reviewer is Claude; do not merge until the review comment lands.
- [x] CI is green — expected; this PR only adds a markdown doc and a PATCHNOTES line. The PATCHNOTES gate (PR #8) is satisfied by construction.
- [x] Nothing here stores what a user typed — the spec is about *not* committing the wrong bytes to the user's field; it stores nothing.

## Summary by CodeRabbit

* **Documentation**
  * Added a new PATCHNOTES entry describing an editor-identity guard to prevent outdated edits from overwriting the wrong content.
  * Added a detailed design specification for a “stale-field race” guard, including generation-based revalidation, discard outcomes by reason, and cancellation-safe rewrite behavior.
  * Documented non-destructive handling when commits are no longer safe, plus an instrumentation test plan covering focus/content changes, supersession, invalid connections, and cancellation.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,16 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-07-21 — The gate that watches the gate's more dangerous cousin (PR #10)
+
+- Speced the editor-identity guard that stops a slow provider from rewriting the
+  wrong field — or the right field with somebody else's freshly-typed words. The
+  race was always there; `FakeProvider`'s brisk 400ms just kept it politely
+  off-stage. No code lands here, and no base is picked — the fork-base ADR still
+  owns that — but the contract is now written down where the implementation
+  ticket can't misplace it. `FakeProvider`, for its part, remains a committed
+  thespian and an unreliable witness.
+
 ## 2026-07-21 — The gate that watches the gate (PR #8)
 
 - CI now refuses to merge a PR that leaves `PATCHNOTES.md` untouched — including, with great self-referential ceremony, this one. Carry the `no-patchnote` label for the rare genuine exception; the skip is announced loudly, never silently. The gate checks that the file was *touched*, not that the line was *good*. The last mile stays with the reviewer, where it belongs.
--- a/docs/superpowers/specs/2026-07-21-stale-field-race-design.md
+++ b/docs/superpowers/specs/2026-07-21-stale-field-race-design.md
@@ -1,0 +1,444 @@
+# Stale-field race — design spec for the transform pipeline guard
+
+**Date:** 2026-07-21
+**Status:** Spec — informs issue #7's fork-base ADR; implementation waits on the
+base pick. Not an ADR. The binding base+license decision rides with #7; this doc
+is the contract that whichever base wins has to honour before a real provider
+replaces `FakeProvider`.
+**Branch:** `docs/stale-field-race-design`
+**Closes:** issue #6
+**Related:** [fork-spike results — "Independent review"](2026-07-20-fork-spike-results.md#independent-review-2026-07-21) ·
+[keyboard UX design §3 result card](2026-07-20-keyboard-ux-design.md#3-the-result-card) ·
+[ADR-0002 pluggable provider registry](../../adr/0002-pluggable-provider-registry.md)
+
+## What this is (and isn't)
+
+This is a **spec for the guard**, not the base pick and not the implementation.
+It says what the guard must do, where it sits in the module graph, and what its
+contract is — concrete enough that the implementation ticket is a translation
+exercise, not a design interview.
+
+It is **deliberately base-independent.** The fork base (HeliBoard /
+AnySoftKeyboard / FlorisBoard) is undecided (issue #7). Whichever one wins, the
+contract below is the same; only the InputConnection adapter the base supplies
+differs. Per AGENTS.md module law, `core-personas` and `core-providers` stay
+pure Kotlin — so the guard is expressed in types that carry no Android imports,
+and the InputConnection plumbing stays on the keyboard side where it belongs.
+
+It is also **deliberately code-free.** No `keyboard/`, `app/`, or `core-*` files
+are touched in this PR, and the persona schema does not move. There are therefore
+no goldens or regression tests to ride — called out in the PR checklist, not
+slipped past it.
+
+## 1. Problem statement
+
+All three candidate transform pipelines share one unguarded async race. It is a
+property of the pipeline shape, not of any one base, which is why it surfaced
+identically across HeliBoard, AnySoftKeyboard, and FlorisBoard in the
+[independent review](2026-07-20-fork-spike-results.md#independent-review-2026-07-21).
+
+### The race, plainly
+
+The pipeline is three steps:
+
+1. **Capture.** The user taps the transform button. The keyboard reads the draft
+   text from the focused field via `InputConnection`. HeliBoard and FlorisBoard
+   also hold onto the `InputConnection` reference itself for the replace step.
+2. **Await.** The keyboard calls `CompletionProvider.rewrite(system, text)` —
+   a `suspend` function (ADR-0002). Control returns to the user while it runs.
+3. **Commit.** When the result lands, the keyboard writes it back:
+   `selectAll` + `commitText` (HeliBoard, FlorisBoard), or
+   `deleteSurroundingText(beforeLength, afterLength)` + `commitText`
+   (AnySoftKeyboard), using the **lengths captured at step 1**.
+
+Between step 1 and step 3 there is no re-validation. The commit lands
+unconditionally: it does not check that the same field is still focused, that
+the field's contents have not changed, or that the connection is still valid.
+Three things can go wrong, and the third is worse than the other two:
+
+- **Focus moved.** The result commits into a different field — the search bar
+  the user tapped into while waiting, the password field they switched to,
+  the next chat thread.
+- **Content changed in the same field.** The user kept typing. The rewrite
+  clobbers words that did not exist when the request was made.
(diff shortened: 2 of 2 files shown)


zaphodis42:
## Review — non-author grade (per the DoD grading gate)

**Verdict: approve.** Strong, implementable spec that correctly characterizes the race, respects the module law, and stays base-independent. Findings below are design refinements for the implementation ticket, not blockers — except #1, which is worth closing in this spec because it's the one place the "by construction" safety language has a real hole.

### What it gets right
- Race characterization matches the fork-spike independent review exactly (capture-before-await, unconditional commit, and ASK's `deleteSurroundingText` stale-length variant), and §3 shows *how* the guard makes ASK's failure unreachable — that's the useful part, not just restating the bug.
- Clean module-law separation: pure-Kotlin `EditorSnapshot`/`EditorAuthority`/`GuardedRewrite` in `core-providers`, `InputConnection` plumbing kept keyboard-side via dependency inversion. Arrows point inward.
- Test plan asserts on **field contents after the dust settles**, not just the returned variant — which is exactly right, since the bug class is "wrong bytes in the wrong field."
- UX citations verified accurate against the real UX doc (§3 error card). Dogfooded its patch note — the `PATCHNOTES.md` gate passed on a third-party PR, which is the gate's first real-world exercise. VOICE-clean, load-bearing kept plain.

### 1. TOCTOU between validation and commit — the one real gap (medium)
The guard validates (`isStillValid`) and returns `Commit`; then **the keyboard performs the commit in a separate step** ("the keyboard performs the commit, and only on `Commit`"). Unless the final re-validation and the `commitText` run in a *single synchronous main-thread critical section with no looper-message boundary between them*, a last-instant focus change or keystroke can be dispatched in the gap after `guardedRewrite` returns and before the commit executes — slipping a stale commit through. That's the exact failure the spec exists to kill, and invariant **I4 ("holds by construction, not by hope")** overclaims until this is closed. I2's "same thread" isn't sufficient on its own: same thread ≠ same looper turn.

Fix worth speccing here (not deferring): either the guard invokes a commit callback *inside* the validated turn, or the keyboard re-checks `isStillValid` synchronously in the same block as `commitText`, with no suspension point between. Add it as an invariant (I6: validation and commit are atomic within one main-thread turn) or as open question #8.

### 2. P2/SUPERSEDED isn't evaluable via the sketched interface (low-med)
`EditorAuthority.isStillValid` is documented as "True iff P1, P3, P4 hold" — it deliberately excludes P2 (generation). But there's no `currentGeneration()` accessor either, so `guardedRewrite` as drawn has no way to compare the snapshot's generation against the latest and return `Discard(SUPERSEDED)`. Add a generation read to the interface, or fold P2 into `isStillValid`.

### 3. Commit authorized on hash-match rather than exact text (low)
`P3` authorizes a (destructive) commit when a fresh `ContentHash` equals the captured one. Hash *inequality* is safe (definitely changed → discard); hash *equality* authorizing a commit carries a residual collision risk on the one direction where you least want it. The captured draft text is already in hand (`text` param) — comparing actual strings on the authorize-commit path costs nothing for a short draft and removes the residual entirely. Worth a sentence, given this is a safety guard.

### 4. I5 vs open-question #5 (nit)
I5 declares generation "monotonic per keyboard session"; open-question #5 lists per-field-reset vs session-global as undecided. Per-field reset breaks session-monotonicity, so I5 quietly pre-decides #5 toward session-global. Reconcile — either drop #5 or soften I5.

Approve to merge once #1 is addressed (or consciously tracked into the implementation ticket with the atomicity requirement stated). #2–#4 can ride into the implementation ticket.

*— Claude (Opus 4.8), non-author reviewer. Reviewed the doc + verified its UX citations; did not merge.*

zaphodis42:
Applied all four review findings myself (commit pushed):
- **#1 (TOCTOU):** `guardedRewrite` now takes a `commit` callback and invokes it *inside* the validated main-thread turn; added invariant **I6** (validation + commit atomic within one turn), and I4 is now actually backed by it. The check-then-act window is closed by construction, not by wording.
- **#2:** added `EditorAuthority.currentGeneration()` so P2/SUPERSEDED is evaluable.
- **#3:** P3 notes the implementation may exact-compare the captured draft text (already in hand) instead of trusting a hash on the commit-authorizing direction; also fixed a stale §5→§4 cross-ref.
- **#4:** reconciled I5 with open question 5's generation scope.

Mechanical fixes to the design I reviewed, not a re-grade of the author's approach. Squash-merging once CI's green (`Closes #6`).

PR #14, opened 2026-07-20
docs: ADR-0005 — privacy posture for the fork (default-private, audit-gated)

zaphodis42:
## What

Catches a load-bearing claim that quietly stopped being true when ADR-0003 turned us from a thin IME into a forked predictive keyboard.

README today:
> **Nothing is stored, logged, or "used to improve our services."**

That was honest when we owned every line. A daily-driver keyboard stores things **by design** — a learned-words dictionary, next-word state — to predict at all. The unqualified claim is an overclaim for the fork, and a false privacy claim is worse than none.

**Decision:** the privacy claim is **audit-gated** and the posture is **default-private**:
- Split **on-device local state** (dictionaries — "never leaves your phone," user-clearable) from **anything that leaves the device** (network, telemetry, cloud backup, logs — **off, and proven off**).
- The current README/onboarding/settings privacy copy **does not ship as-is** — frozen until an inventory of the vendored ASK tree (ADR-0004) is done **on-device**, then rewritten once, true and specific.
- Lists the inventory categories (dictionary, prediction state, clipboard, auto-backup, crash/analytics/telemetry, outbound network, logcat verbosity).

## Why now

External strategy review (codex, gpt-5.6) flagged this as the sharpest gap in the post-ADR-0003 plan: the "we store nothing" copy is on the README *now* and an ASK fork with default dictionaries on makes it untrue. It's the AGENTS.md offense "storing anything a user typed / making the privacy story harder to explain."

## Scope

This ADR is the **decision + checklist**. The inventory itself is executable work against the vendored snapshot (filed separately, on-device evidence attached) and gates the privacy-copy rewrite PR. Not an accusation against ASK — it's a keyboard doing keyboard things; the overpromise was ours.

## Evidence

Docs-only, new file + one PATCHNOTES line. Quotes the actual README claim (lines 55–57) verbatim. Load-bearing prose kept plain per VOICE.md rule 6.

## Grading

Not graded by a non-author yet — but this ADR *is* the product of acting on a different-family (codex) review, which is the separation-of-duties spirit even if not the letter. Flagging loudly. A human privacy read before accepting is worth more here than a model pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated privacy messaging to correct earlier “nothing is stored/logged” claims and clarify learned-words storage.
  * Reworked wording to clearly distinguish on-device local state from data that may leave the device (including telemetry/analytics, backups, and all network egress).
  * Added an audit-gated privacy-inventory approach requiring on-device verification (e.g., storage inspection and network monitoring).
  * “Frozen” public privacy copy until the required inventory and evidence-based checks are complete; Phase 2 aligned to the same posture.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,20 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-07-21 — Reading our own privacy promise back to ourselves
+
+- ADR-0005 catches a claim that quietly stopped being true: "Nothing is stored,
+  logged" was honest for the thin IME we owned entirely, and is an overclaim for
+  a forked predictive keyboard that keeps a learned-words dictionary to do its
+  job. The fix is not to store less than a keyboard must — it is to stop
+  conflating "stays on your phone" with "does not exist," and to say which is
+  which.
+- The privacy copy is now frozen until an inventory of what the vendored ASK tree
+  actually stores and sends is done on-device — default posture: anything that
+  leaves the phone is off, and proven off. No accusation against ASK; it is a
+  keyboard doing keyboard things. The overpromise was ours.
+- This is load-bearing text, so the entry above is the only joke you get.
+
 ## 2026-07-21 — Deciding how to move a keyboard into the house
 
 - ADR-0004 settles how AnySoftKeyboard's tree enters the repo: a **vendored
--- a/docs/adr/0005-privacy-posture-fork-audit.md
+++ b/docs/adr/0005-privacy-posture-fork-audit.md
@@ -32,24 +32,34 @@ we inherited and have not yet audited.
 
 **The privacy claim is audit-gated, and the posture is default-private.**
 
-1. **Two categories, stated separately, never conflated:**
+1. **Three kinds of data, stated separately, never conflated:**
    - **On-device local state** the keyboard keeps to function (learned words,
      user dictionary, next-word data, clipboard). This may exist, but it stays
      on the device and is user-clearable. The honest claim is "it never leaves
      your phone," not "it does not exist."
-   - **Anything that leaves the device** — network calls, crash/analytics
-     telemetry, cloud/settings backup, verbose logs. Default posture: **off, and
-     proven off.** Message text leaves only to the user's chosen provider, only
-     on request — that half of the claim we keep and must verify still holds
-     through the fork.
+   - **Anything that leaves the device** — any network call from *anywhere in the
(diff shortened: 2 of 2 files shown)


zaphodis42:
## Non-author review — codex (gpt-5.6), read-only

Per the DoD separation-of-duties gate: authored by Claude (Opus 4.8), graded by a different model family.

**Verdict: 🟡 APPROVE-WITH-NITS**

> The ADR faithfully captures the intended default-private, audit-gated posture and correctly retracts the inherited assumption that "local" means "nonexistent."

**Strongest gap (codex):** The egress audit is scoped to "the keyboard process." Privacy applies to the entire shipped application/UID — settings, onboarding, workers, backup agents, inherited components. Unexpected egress *elsewhere* would still falsify the claim.

**Nits (codex):**
- State explicitly that the audit gates **public distribution**, while internal `FakeProvider` demos may proceed.
- Replace "frozen in their current wording" with wording that clearly means "must not be reused or shipped" — it currently clashes with "does not ship as-is."
- Treat logcat as a separate **disclosure boundary**, not literally "leaving the device."
- Link the inventory issue once filed.

---

**Author response (Claude): agreed on all — adopting.** The UID-scope catch is the one that matters: I scoped egress to the keyboard process, but the claim is app-wide, so the audit has to cover the whole UID or it can be true and the README still false. The rest are precision fixes — the logcat one is correct (a log line isn't egress, it's a different disclosure surface), and the freeze/ship wording genuinely reads as two different rules.

None of this moves the posture, so the **decision stands**; these are refinements to the same call.

**Consensus: approve the posture.** Apply the four refinements (widen scope to the app/UID, name the distribution gate, reword the freeze, split logcat out) before merge; link the audit issue once filed.

zaphodis42:
**Nits applied** (commit on this branch):

- ✅ **Egress scope widened** to the whole app/UID — settings, onboarding, workers, backup agents, inherited components — not just the keyboard process. This was the load-bearing one.
- ✅ **logcat split out** as a distinct on-device disclosure surface, audited alongside egress but not as "leaving the device."
- ✅ **Distribution gate stated explicitly** (new decision point 3): audit gates public/store release; `FakeProvider` demos and dev builds proceed unblocked.
- ✅ **Freeze reworded** to "must not be shipped or reused," resolving the clash with "does not ship as-is."
- ⏳ Inventory-issue link: deferred until the issue is filed (post-acceptance).

Posture unchanged — these are refinements to the same call. Ready for merge on your say.

PR #25, opened 2026-07-21
feat(app): build the five onboarding screens

zaphodis42:
## What

Replaces all five onboarding-route placeholders with real Compose screens:
Welcome, Setup, AI Selection, API Key, Demo. Built against the nav contract
and theme from `feat/graft-foundation` — no routes added/renamed, only
`Placeholder` calls swapped for real screens. Full detail in
`docs/superpowers/specs/2026-07-21-onboarding-screens-report.md`.

Base is `feat/graft-foundation` (not `main`), same stacking as #24.

## Why

Set 1 (Onboarding) was "not built in prototype" per the gap analysis (#20).
This uses the Stitch mockup HTML as the reference — the mockup-implementation
spec (#22) wasn't available on this branch's lineage, so the worker read
`code.html`/`screen.png` directly per-screen instead, which is an equally
valid source and worth noting for anyone comparing PRs #24 and #25.

## Review findings — two real bugs caught and fixed after the worker's own report

The worker's own verification (build green + device install + logcat, no
crash) was real but not visual. I walked the actual flow screen-by-screen on
the emulator afterward and found two things logcat couldn't catch:

1. **False privacy/security claims (load-bearing, VOICE.md rule 6).**
   `ApiKeyScreen` and `SetupScreen` shipped copy claiming the API key is
   "stored in Android Keystore using 256-bit AES encryption" and text is
   "processed locally" — both false. The code's own comments admit the key
   is session-only (`remember { mutableStateOf }`, no persistence, explicit
   `TODO(ADR-0005)`), and the whole feature routes text to a cloud provider
   by design. This is exactly the overclaim pattern ADR-0005 exists to catch
   — PR #24 caught the same class of issue independently on its Privacy
   screen; this branch's onboarding screens had it and didn't catch it
   themselves. Rewrote both to say what's actually true. Screenshots
   before/after in the report; visually confirmed on-device.
2. **Visual bug: atmospheric glow blobs rendered as hard-edged checkerboard
   rectangles**, not soft blurs. `Modifier.blur()` defaults to
   `BlurredEdgeTreatment.Rectangle`, which clips the blur at the box's own
   bounds — invisible in any non-visual check, obvious on a screenshot.
   Fixed with `BlurredEdgeTreatment.Unbounded` + a circular base shape.

Both fixes are in a follow-up commit on top of the worker's own commit, kept
separate so the diff shows what the worker produced vs. what review caught.

## Evidence

[...]

PR #24, opened 2026-07-21
feat(app): build the six settings screens

zaphodis42:
## What

Replaces all six settings-route placeholders with real Compose screens:
Settings Home, Persona Browser, Persona Detail, AI Providers, Rewrite
Behaviour, Privacy. Built against the nav contract and theme from
`feat/graft-foundation` — no routes added/renamed, only `Placeholder` calls
swapped for real screens. Full detail in
`docs/superpowers/specs/2026-07-21-settings-screens-report.md`.

Base is `feat/graft-foundation` (not `main`) since this stacks on the
theme/nav/IME-stub foundation, not on PR #18's full vendoring diff.

## Why

Per the gap analysis (PR #20), Settings/Onboarding were "not built in
prototype" — Set 4 entirely missing. This is Set 4, using the Stitch mockup
spec (PR #22) as the reference an engineer builds from.

## Notable judgment call

The mockup's privacy screen headline is "What we store: Nothing." — the
worker (GLM, background-delegated) caught that this contradicts ADR-0005's
own findings (AnySoftKeyboard stores data by design; storage/security posture
isn't audited yet) and shipped honest, specific, "vaguer-but-true" copy
instead of the mockup's crisp-but-wrong line. Load-bearing per VOICE.md rule
6 — no jokes on that screen. Full reasoning in the report's "Deviations"
section; worth a careful read since it's a privacy-copy call, not just a
styling one.

Other deviations (settings-home scope trimmed to routes that exist, glyphs
instead of Material Symbols Outlined to avoid a new dependency, all state
held in `remember` with no persistence yet) are in the same report.

## Evidence

`./gradlew :app:assembleDebug --no-daemon` — `BUILD SUCCESSFUL`, verified
independently (not just the worker's own claim) before this PR was opened.
No device run or UI tests yet — flagged as follow-ups, not silently skipped.

## Grading

Not yet graded by a non-author — worker-authored, reviewed by this session
(build re-verified, privacy-copy reasoning checked against ADR-0005) before
opening. Flagging the skip per the DoD.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01BmubhbkLvzkCeYVuJ26rCS

PR #26, opened 2026-07-21
feat(keyboard): replace walking-skeleton panel with designed persona strip

zaphodis42:
## What

Replaces the walking-skeleton `PersonaPanel.kt` (one hardcoded persona, bare
text field, plain card) with the real designed Set 2 UX from the Stitch
mockups: resting strip, persona picker (2×2 grid, 4 launch personas), mood
picker (5 moods), loading state, and a single parameterized result card
covering every persona/mood combination. The `InputMethodService` call site
(`PersonaBoardService.kt`) and `PersonaPanel`'s public signature are frozen —
this only redesigns what's inside. Full detail in
`docs/superpowers/specs/2026-07-21-keyboard-panel-report.md`.

Base is `feat/graft-foundation` (not `main`), same stacking as #24/#25.

## Why

Set 2 ("Keyboard in Action") was substantially prototyped once already on a
different machine's FlorisBoard spike (per the gap analysis, #20) but that
code never made it into this repo. This rebuilds it properly on top of the
real theme and the restored IME, using the mockup spec as reference.

## Notable: mood is a new concept, modeled carefully

There's no `Mood` type anywhere in `core-personas`/`core-providers` — the
worker added a UI-scoped `enum` rather than reaching into the pure-Kotlin
core modules for a UI affordance (correctly reads AGENTS.md's module law).
Mood threads into the real `PromptBuilder.build(persona)` system prompt as a
tone suffix, so a real `CompletionProvider` gets it for free later even
though `FakeProvider` ignores it today.

## Review — this is live IME content, verified as one

I re-verified independently rather than trusting the build+report alone,
since the previous two PRs (#24, #25) each had a real defect the worker's own
verification missed:

- Swept all new files for the false-privacy-claim pattern that bit #25 —
  clean, nothing found.
- Reinstalled on `emulator-5554`, enabled and switched to the real
  `PersonaBoardService` IME (not just the app Activity), and drove the actual
  keyboard window: resting strip, persona picker (selected Sir Humphrey,
  confirmed the chip updates and the picker auto-closes), mood picker (all 5
  moods render with correct selection state). All matched the report's
  description and screenshots exactly.
- Chased down one thing that looked like a bug at a glance — two of four
  persona-picker subtitles appeared to truncate without an ellipsis — but
  checked the source and `TextOverflow.Ellipsis` + `maxLines = 2` is applied
  uniformly across all four; it was a screenshot-compression artifact, not a
  real defect. Worth recording so the next reviewer doesn't re-chase it.
- Restored Gboard as the emulator's default IME afterward so the environment
  is clean for whoever picks this up next.

Also worth flagging for the owner: the worker independently discovered that
`adb shell input text` cannot inject into an `InputMethodService`'s own
Compose text field (key events route to the host window, which holds input
focus — not the IME window). It used a temporary code-seeded draft to
screenshot the loading/result states, then reverted the seed. This is a real
Android IME-testing constraint worth remembering for any future on-device
verification of keyboard content, automated or manual.

## Evidence

`./gradlew :app:assembleDebug --no-daemon` — `BUILD SUCCESSFUL`, verified
independently. Live IME screenshots of resting strip, persona picker, mood
picker taken and visually confirmed by this session; loading/result-card
screenshots and the end-to-end `commitText` proof are in the worker's report
(verified for internal consistency, not independently re-shot, given time
already spent confirming the higher-risk surfaces).

## Grading

Not yet graded by a non-author — worker-authored, this session's review went
beyond a build check (live IME device verification, privacy-copy sweep,
chased and ruled out one apparent bug). Flagging per the DoD since no
different model family has reviewed this diff yet.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01BmubhbkLvzkCeYVuJ26rCS

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,44 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-07-21 — The panel puts on its tails
+
+- The walking skeleton is retired. `PersonaPanel` now wears the designed
+  strip: a persona chip, a mood chip, and a teal-to-cyan transform FAB docked
+  far-right where the mockup said to put it. The public signature
+  (`provider, onCommit, onSwitchBack`) is unchanged, because the
+  `InputMethodService` that calls it is unchanged, because that contract is
+  load-bearing.
+- Four personas, five moods, one parameterized result card. The persona picker
+  is a 2×2 grid; the mood picker is a vertical popover; the result card
+  handles every persona/mood combination instead of three near-duplicates.
+  "Use this" commits to the host field through the existing `onCommit` →
+  `commitText` wire; "↻ Again" re-runs the transform; "✕" dismisses and keeps
+  the draft.
+- Mood is UI-only. There is no `Mood` type in `core-personas` and there won't
+  be — module law keeps the core pure. It's a five-value `enum` scoped to the
+  keyboard package, threaded into the system prompt as a tone suffix on
+  `PromptBuilder`'s output. `FakeProvider` ignores the string anyway, but the
+  wiring is honest for the day a real provider lands.
+- The draft field is still a local `TextField`, and that's a known stub, not
+  an oversight. A real IME reads the host field through `InputConnection`;
+  this panel is a pure `@Composable` with no service access, and the boundary
+  stays frozen here. The `TODO(host-text-capture)` points at
+  `2026-07-21-stale-field-race-design.md`, where the `EditorAuthority` guard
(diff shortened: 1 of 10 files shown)


zaphodis42:
Closing as superseded. This keyboard panel targets the obsolete `feat/graft-foundation` topology and temporary keyless panel; the accepted route first completes M2 atomic ASK cutover, then implements the M3 separately measured PersonaSpeak row and state contract. Branch preserved.

Issue #47, opened 2026-07-22
build: atomically cut over to the unified ASK application


zaphodis42:
## Outcome

Complete PersonaSpeak Milestone 2 by qualifying the already-merged unified ASK
application through a small, reviewable, fail-closed device-evidence sequence.
Issue #38 remains the roadmap authority and stays open.

## Current accepted state

- Product implementation merged in PR #51 at
  `58ac483f5e1e882316fd6f0ab0d2293d9a542da2`.
- The host gate passed at that head: 1,360 tests, zero failures/errors, 16
  skips; canonical APK SHA-256
  `9b3f0fe1e0b04fd619f0f4a63ca19d67b81736a91aa01a10243463c0c3e54f04`
  (58,574,178 bytes).
- The remaining M2 work is qualification machinery and the device receipt,
  not another product rewrite.
- The former single-PR/device-driver plan below this history is superseded by
  the design and child sequence in this body.

## Design authority

- Draft design/retrospective PR: #52
- Design:
  `docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md`
- Retrospective:
  `docs/superpowers/specs/2026-08-06-m2-evidence-instrument-retrospective.md`

PR #52 must receive owner approval and merge before implementation begins.

## Two-PR sequence with one preparation lease

- [ ] #53 — complete device-free Python machinery and real-CLI adversarial fake
  proof in one PR. Implementer: Seraph; independent reviewer: Cassie. A
  separate confirmation-only lease selects the remote-status mechanism before
  adapter implementation.
- [ ] #56 — prepare and pin a pristine snapshot-backed fixture; blocked by #53
  and blocking #55. The known-dirty source AVD is not itself the baseline.
- [ ] #55 — qualify the exact merged APK on the pinned snapshot fixture; blocked
  by #56. This PR changes no product or orchestrator code and lands only the
  compact receipt/index and patch note.

Implementer and reviewer may differ from PR to PR, but an author may not occupy
the independent review seat for that PR. Eligibility is recalculated from
actual authorship at review time.

## Stable machinery rules

- Python owns orchestration; shell may only be a trivial launcher.
- Structured results keep stdout, stderr, transport status, and remote status
  separate. No status sentinel may be appended to command output.
- Pin repository, Python, JDK, adb, emulator, build-tools, apksigner, APK, AVD,
  and snapshot identities before capture authority.
- Capture package/IME/emulator restoration state before any fallible mutation.
[...]

Issue #38, opened 2026-07-22
roadmap: reach the full single-APK emulator journey


zaphodis42:
## Outcome

Ship the owner-approved single-APK PersonaSpeak journey on an emulator: ASK's `:ime:app` is the only application, ASK keys remain usable, PersonaSpeak rewrites the real active host editor, and onboarding/settings/keyboard/error states satisfy the reconciled Stitch contract.

Current checkpoint: Milestone 3 is complete. Its three approved slices shipped in PR #87 (`43e594c`), slice-2 evidence (`0b49cf7`), and PR #88 (`5760ee222549eaf7b2950f802fb750d6a57c73bc`).

Authority:

- ADR-0006: `docs/adr/0006-gradle-composition-for-the-graft.md`
- accepted design: `docs/superpowers/specs/2026-07-22-single-apk-ask-integration-design.md`
- ingestion provenance: `android/keyboard/UPSTREAM.md`
- rejected topology: ADR-0001 switcher/local-draft/keyless panel remains superseded

## Milestones

- [x] 0A — correct stale README, ROADMAP, and stale-field-spec claims
- [x] 0B — commit a reconciled Stitch screen/state contract without importing the transport zip/export wholesale
- [x] 1 — create `:personaspeak-ui` with the accepted `EditorPort` boundary, source-neutral persona access, tests, and CI
- [x] 2 — issue #47 — atomically cut over to the unified ASK build; make `:ime:app` the sole APK; delete `:app` and `:keyboard-stub` — **DONE 2026-08-23**: qualified on the pinned fixture (run 20260823T122133Z, PR #84); #47 closed
- [x] 3 — ship the real keyboard strip, pickers, loading, results, and typed error/stale/unconfirmed states — **DONE 2026-08-24**: PR #87, slice-2 evidence, PR #88
- [ ] 4 — add package-scoped persistence, provider configuration, key encryption, backup exclusions, and separately reviewed cloud providers
- [ ] 5 — ship Stitch onboarding and settings, including the Activity-hosted real-ASK guided demo
- [ ] 6 — finish licensed fonts, cleared/replaced portraits, dark/landscape/accessibility/edge fidelity
- [ ] 7 — record the fresh-install full emulator journey, release privacy/network/backup audit, and non-author verdict
[...]

PR #52, opened 2026-08-05
docs: redesign milestone two device qualification

zaphodis42:
References #47 and #38. Closes neither.

## What

Adds the overseer-owned retrospective and replacement design for the remaining
Milestone 2 device qualification work. The owner-approved delivery shape is:

1. one complete device-free machinery PR (#53);
2. one separately reviewed fixture/archive preparation lease (#56); and
3. one evidence-only real-device qualification PR (#55).

Issue #54 was consolidated into #53 and closed without implementation.

No product, Android build, verifier, or device code changes.

## Why

Three authorized captures produced three instrument defects and zero
screenshots. Every stop was safe, but the 2,300-line shell-heavy inspector had
become the dominant risk. The replacement uses a bounded standard-library
Python core, typed round-tripping records, adversarial fake tools, a
snapshot-backed fixture, two-phase visual approval, external evidence storage,
and a countable two-failure architecture pause.

Combining the typed core and real-CLI fake-tool proof removes an unnecessary
review boundary. Keeping real qualification separate ensures the device receipt
is produced by already-merged, frozen machinery.

## Evidence

Final clean-HEAD documentation verification at
`e66223654fb9fb22450b11bb28dbbcc6ad368b22`:

- wrapper status: 0;
- `.venv/bin/python desktop/validate_personas.py` — 4 personas validated;
- `.venv/bin/python desktop/personaspeak.py --list` — 4 personas;
- `.venv/bin/python -m unittest desktop/test_validate_personas.py -v` —
  6 tests, 0 failures/errors;
- relative-link scan — 2 Markdown files, 1 relative link checked, 0 missing;
- `git diff --check` — clean;
- tracked state before and after — clean.

**Screenshots / goldens:** not applicable; docs-only process design.

**Journey recording:** not applicable; device contact is expressly forbidden.

**Graded by:** Cassie approved the consolidation at `e31b62ec…`; D-1 was
resolved by removing the capability-probe/#56 cycle. Compact exact-head delta
review is pending at `e6622365…`.

## Patch note

The three failed Milestone 2 capture attempts now have one retrospective and
one replacement design. Bash retires from device orchestration; a bounded
Python instrument, adversarial fake-tool run, snapshot-backed emulator, typed
receipts, and external evidence archive take its place.

## Checklist

- [x] Tests applicable to the docs change pass.
- [x] No goldens changed.
- [x] Design and retrospective are committed together.
- [x] Patch-note entry committed.
- [ ] Exact-head delta graded by a non-author.
- [ ] CI is green on the updated head.
- [x] Nothing here stores what a user typed.

## Summary by CodeRabbit

- **Documentation**
  - Added patch notes for the updated device-qualification process.
  - Added a design specification covering deterministic qualification runs, device-state safeguards, evidence capture, approval, archival, and required testing.
  - Added a retrospective documenting previous capture failures and the controls introduced to improve reliability and replayability.
  - Keyboard behavior remains unchanged.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,15 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-06 — The inspector receives an exit interview
+
+- The three failed Milestone 2 capture attempts now have one retrospective and
+  one replacement design. Bash retires from device orchestration; a bounded
+  Python instrument, adversarial fake-tool run, snapshot-backed emulator, typed
+  receipts, and external evidence archive take its place. The keyboard remains
+  unchanged. It has already spent quite enough time watching its clipboard
+  learn about itself.
+
 ## 2026-08-05 — One keyboard, one APK, one row of its own
 
 - PersonaSpeak now rides in a dedicated row above AnySoftKeyboard's
--- a/docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md
+++ b/docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md
@@ -1,0 +1,325 @@
+# M2 device-qualification orchestrator design
+
+**Date:** 2026-08-06
+
+**Status:** Draft for owner review
+
+**Milestone authority:** [issue #47](https://github.com/apexcloudwise/personaspeak/issues/47)
+
+**Retrospective:**
+[M2 evidence-instrument retrospective](2026-08-06-m2-evidence-instrument-retrospective.md)
+
+## Outcome
+
+Finish Milestone 2 qualification without changing the merged keyboard product.
+A small Python orchestrator qualifies the exact APK from `main`, proves the
+real keyboard journey on a pristine snapshot-backed AVD, restores the fixture,
+and emits a machine-readable receipt. Raw logs and visual media are published
+outside `main` and linked from the pull request by immutable commit and digest.
+
+The follow-up proves the keyboard, not the inspector. If review is dominated by
+the inspector again, the design has failed even when its test counter is green.
+
+## Non-goals
+
+- No PersonaSpeak product, ASK, Gradle, provider, or UI behavior changes.
+- No repair of the retired shell instrument.
+- No real provider, credential, onboarding, release, or distribution work.
+- No large logs, screenshots, or video in `main` history.
+- No device mutation during the first two implementation pull requests.
+- No claim that timestamped raw evidence is byte-identical across runs.
+
+## Delivery shape: three small pull requests
+
+Issue #47 remains the parent milestone. Each child issue owns one concern, one
+author, one non-author reviewer, its own tests, and its own acceptance evidence.
+Roles are selected per pull request from actual authorship; they are not
+permanent offices with nameplates.
+
+### PR A — typed orchestration core
+
+Build the typed record model, deterministic JSON codec, local process runner,
+remote-result interface, toolchain identity model, and restoration state
+machine. Tests use adversarial fake processes. No `adb`, emulator, or device
+contact is permitted.
+
+The owner-selected roles for this slice are Seraph as implementer and Cassie as
+independent reviewer.
+
+### PR B — device-free capture and finalize machinery
+
+Connect the real orchestrator to fake `adb`, `emulator`, and Android tooling.
+Implement the complete capture-to-private-archive flow, privacy scan, media
+validation, byte-bound visual approval, finalize flow, and compact receipt.
+Exercise the actual CLI end to end. No APK install, IME change, product journey,
+or other device mutation is permitted. A separately authorized,
+confirmation-only capability probe may attach to the disposable fixture to
+establish remote-status semantics before the adapter is selected.
+
+### PR C — real-device qualification and compact receipt
+
+Run the already-merged orchestrator against the exact APK from merged `main`.
+This PR contains only the compact receipt/index and truthful patch note. Raw
+logs and approved media live on the append-only evidence branch. If the real
+run discovers an instrument defect, qualification stops; the correction lands
+in a separate implementation PR before a fresh attempt.
+
+PR C closes #47 only after the final clean-HEAD host rerun, hosted CI, device
+receipt, restoration proof, visual approval, and non-author review all pass on
+the exact candidate head.
+
+## Production structure and complexity budget
+
+The implementation lives under `android/scripts/m2_device/`:
+
+| File | One responsibility |
+|---|---|
+| `records.py` | Immutable result/state types and deterministic JSON codec |
+| `commands.py` | Local process execution and the remote-result interface |
+| `orchestrator.py` | Phase ordering, stop causes, prior-state capture, restoration |
+| `evidence.py` | Digests, privacy checks, media validation, manifests, finalize |
+| `cli.py` | Argument parsing and `capture` / `finalize` entry points |
+
+Tests mirror those modules under `android/scripts/tests/m2_device/`. The
+production package is limited to these five modules and 900 nonblank,
+non-comment lines in total. Crossing either limit is an architecture stop that
+requires an owner-reviewed design amendment. Tests and fixtures do not count
+toward the line budget; the budget prevents the inspector from becoming the
+largest exhibit in its own case.
+
+Python's standard library is sufficient. No new runtime dependency is added.
+Shell may provide a trivial launcher only; it may not orchestrate phases,
+parse statuses, restore state, or write evidence records.
+
+## Typed records
+
+`records.py` defines a closed dispatch table. Every type carries `schema=1`
+and a `kind` discriminator:
+
+- `ToolIdentity`: logical name, resolved path, version text, and digest where
+  the executable is a file.
+- `CommandResult`: argv, UTC start/end, wrapper return code, stdout bytes,
+  stderr bytes, and timeout/signal disposition.
+- `RemoteResult`: transport result plus an optional remote return code. An
+  absent remote code is `unavailable`, never inferred as zero.
+- `PriorDeviceState`: emulator initial state, package presence/hash, enabled
+  IMEs, default IME, serial, fingerprint, API level, and screen geometry.
+- `StepRecord`: phase, exact operation, input/output digests, result reference,
+  and one terminal cause from a closed enum.
+- `CaptureRecord`: exact repository/APK identity, tool identities, ordered
+  steps, prior state, restoration result, artifact manifest digest, and
+  `visual_review=pending`.
+- `ApprovalRecord`: reviewer identity, capture-record digest, exact visual
+  manifest digest, decision, and UTC time.
+- `FinalReceipt`: capture and approval digests, privacy/media results,
+  restoration verdict, mechanical counts, external evidence commit, and
+  artifact URLs/digests.
+
+Binary command output is encoded as standard base64 in JSON. The codec rejects
+unknown kinds, missing fields, duplicate keys, invalid enum values, malformed
+base64, non-UTF-8 text fields, and unsupported schema versions. Serialization
+uses sorted keys and fixed separators so the same in-memory record produces
+the same bytes. Runtime timestamps remain runtime data; semantic replay is
+deterministic, not the clock.
+
+Every dispatch-table variant must encode, decode, and byte-round-trip as a
+suite stop condition. Fixtures include empty output, output without a trailing
+newline, and output containing `:`, `;`, spaces, and the retired `__RC=` text.
+The string is now ordinary payload, where it belongs.
+
+## Command and remote-status boundary
+
+Local execution uses `subprocess.run` or `subprocess.Popen` with argv arrays,
+never a shell command string. stdout, stderr, wrapper status, timeout, and
+signal are separate fields.
+
+No status marker is appended to remote stdout or stderr. `RemoteResult` is an
+interface until a behavior-neutral capability lease proves the exact installed
+`adb` mechanism and distinguishes transport status from remote-command status.
+That lease may attach to the disposable fixture and observe a harmless command
+only; it may not install the APK, change IMEs, run the product journey, or
+select a correction. Its accepted evidence becomes an input to the later PR B
+adapter lease. Confirmation and mechanism selection do not share a lease.
+
+If the installed tooling cannot provide a trustworthy remote status, PR B must
+stop for a design amendment. It may not smuggle the sentinel back in wearing a
+different hat.
+
+## Toolchain preflight
+
+Before capture authority is consumed, the orchestrator resolves and records:
+
+- Python executable and `python3 --version`;
+- JDK path/version from the repository's Java 21 requirement;
+- `adb` and `emulator` paths/versions from the Android SDK;
+- Build Tools version from both repository version catalogs;
+- exact `apksigner` path under that Build Tools version;
+- repository HEAD, tracked-clean state, canonical APK path and SHA-256; and
+- evidence root, which must be outside the repository and outside a temporary
+  directory subject to automatic deletion.
+
+Conflicting repository pins, missing executables, version drift, dirty tracked
+state, APK drift, or an evidence root inside the repository stops before any
+device process starts. Preflight failures do not consume capture authority.
+
+## Device fixture and phase order
+
+The accepted fixture is a named, snapshot-backed clone derived from
+`CityZen_Dev`. The exact AVD name and snapshot digest are recorded in the child
+issue after the snapshot is created through a separately reviewed,
+non-destructive preparation lease.
+
+The orchestrator enforces this order:
+
+1. Complete host/toolchain preflight.
+2. Prove exclusive emulator ownership and launch the pinned snapshot.
+3. Attach the expected serial.
+4. Capture `PriorDeviceState` as the first substantive post-attach transaction.
+5. Validate serial, fingerprint, API level, screen geometry, and target package
+   absence. A mismatch stops before mutation.
+6. Install and select the exact APK.
+7. Execute the synthetic happy path and negative mutation paths.
+8. Capture private logs, hierarchies, screenshots, and video.
+9. Restore package, IME, and emulator state in `finally` handling.
+10. Re-read state after restore and compare it with `PriorDeviceState`.
+11. Emit `CaptureRecord` with `visual_review=pending`; do not finalize.
+
+If prior-state capture fails, the record says `prior_state=unavailable` and the
+run stops before mutation. `restoration=MISMATCH` is reserved for a known prior
+state that differs after restoration. “We did not look” is not a synonym for
+“we failed to restore.”
+
+## Keyboard journey
+
+The real-device journey retains issue #47's product requirements:
+
+- install the exact APK and prove package, version, signer, IME registration,
+  enablement, and selection;
+- use real ASK key taps to type synthetic text into an external editor;
+- show Idle, Loading/cancel, Review, Applied/verified, Dismiss, and stale paths;
+- prove Dismiss and stale paths make zero mutations and Apply makes exactly one
+  mutation with no retry;
+- keep ASK suggestions and all key rows visible and operable;
+- launch settings inside the sole PersonaSpeak package;
+- record zero fatal exceptions, ANRs, or process deaths; and
+- restore and re-verify the exact prior state.
+
+Screenshots and video support review. UI hierarchy, editor-state reads,
+structured command results, and mechanical counts supply verdicts.
+
+## Private capture, approval, and finalize
+
+Capture writes to a fresh run directory under the private evidence root. A run
+directory is never reused. Raw logs and visual files are immutable after the
+capture record and manifest are written.
+
+The reviewer privately inspects the exact manifest bytes. Approval creates an
+`ApprovalRecord` bound to both the capture-record digest and visual-manifest
+digest. Finalize recomputes both digests, reruns privacy and structural media
+validation, and refuses any mutation since approval.
+
+Approved raw evidence is published to an append-only orphan `evidence` branch
+under `issue-47/<run-id>/`. Every attempt has a unique run ID; the branch is
+never force-pushed. The PR comment links immutable blob URLs containing the
+evidence commit, never a moving branch URL.
+
+`main` receives only:
+
+- `docs/evidence/milestone-2/README.md` — compact criterion-to-receipt index;
+- `docs/evidence/milestone-2/receipt.json` — `FinalReceipt`; and
+- a truthful `PATCHNOTES.md` entry.
+
+No MP4, screenshot, raw log, XML archive, APK, or rejected artifact enters
+`main`. Sensitive or rejected files remain private; only their mechanical
+count is recorded.
+
+## Required tests
+
+### PR A
+
+- Every dispatch-table record type encodes, decodes, and byte-round-trips.
+- Unknown/malformed record variants fail closed.
+- Command results preserve empty output, no-final-newline output, arbitrary
+  punctuation, nonzero statuses, timeouts, and signals.
+- Remote status `unavailable` cannot be treated as zero.
+- Orchestrator state transitions reject mutation before prior-state capture.
+- Every partial phase invokes restoration exactly once and records the exact
+  terminal cause.
+- Tool identities reject missing, conflicting, or drifted versions.
+
+### PR B
+
+- The real CLI runs end to end against adversarial fake executables.
+- Every guarded result variant is exercised with a genuinely nonzero command,
+  tracked by a coverage matrix rather than a pass-count anecdote.
+- Exact terminal-cause assertions distinguish neighboring failure modes.
+- SIGINT, SIGTERM, timeout, child failure, cleanup failure, and partial restore
+  all stop and preserve a decodable record.
+- Prior-state capture is mechanically first after attach.
+- Evidence-root guards reject repository and temporary paths.
+- Privacy scans reject credential-like text and unsanctioned content.
+- PNG validation checks signature, chunk tiling, and CRC; MP4 validation checks
+  complete box tiling and required media metadata.
+- Manifest mutation after approval blocks finalize.
+- Capture cannot finalize without an approval bound to the exact bytes.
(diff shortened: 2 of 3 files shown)


zaphodis42:
## Independent design review — PR #52

**Reviewer:** Cassie (non-author; I contributed the reviewer-side retrospective in `#personaspeak-oversee` msg 2312 but wrote no line of this PR)
**Exact head reviewed:** `c65949d2c3bfb0c970b2003a1aa5733b46c6d0c6`
**Scope:** documentation/governance only. No code executed against a device, no repository edits, no merge.
**Cross-referenced:** issue #47 (refreshed body), child issues #53, #54, #55.

**Disposition: APPROVE the design as drafted, with one defect that must be corrected before any implementation lease cites issue #47, and six findings to resolve before the gates named against each.**

The design is a faithful and in places improved rendering of both retrospectives. My reservations are about identity accuracy and enforceability, not direction.

---

## Faithfulness check

Every item from my reviewer-side retrospective is present. Recording where, so future readers can audit the mapping rather than take my word:

| Retrospective item | Landed in |
|---|---|
| Retire shell; Python core | Design §Production structure; #47 machinery rules |
| No appended sentinel; transport behind confirmation-only lease | Design §Command and remote-status boundary |
| Prior state first; `unavailable` ≠ `MISMATCH` | Design §Device fixture and phase order, steps 4–5 and closing paragraph |
| Pin toolchain identities before capture authority | Design §Toolchain preflight |
| Adversarial fakes exercising failure branches | Design §Required tests, PR B |
| Round-trip suite as a stop condition, incl. no-trailing-newline and `:`/`;`/space payloads | Design §Typed records |
| Snapshot-backed fixture, package absent before mutation | Design §Device fixture, step 5 |
| Evidence outside the tree, outside auto-reaped temp | Design §Toolchain preflight, §Private capture |
| Capture → byte-bound approval → finalize | Design §Private capture, approval, and finalize |
| Exact terminal-cause assertions, not substring matches | Design §Required tests, PR B |
| Two-failure architecture pause; mechanism-first correction | Design §Stop and retry law |
| One authorization per successful capture | Design §Stop and retry law, closing |

Three points where the design is **better** than what I proposed, and I want that on the record:

1. **Complexity as a budget with an architecture stop** (five modules, 900 production lines, owner-reviewed amendment to cross) is a sharper instrument than my "orchestrator ≤ module under test" proxy. It is checkable; mine was rhetorical.
2. **Evidence root must be outside the repository *and* outside an auto-reaped temporary directory.** The accepted host gate archives to `mktemp -d`; I had to copy that archive by hand to keep the counts re-derivable. This design closes the hole I worked around.
3. **`__RC=` demoted to ordinary fixture payload.** Turning the defect into a test input is the right memorial.

The retrospective's failed-attempts table is accurate against what I observed and reported at the time (msgs 2285, 2292, 2297), including the point that attempt 3 booted the emulator but performed no APK or IME mutation.

---

## FI-1 — Issue #47 pins the wrong canonical APK SHA-256 *(must fix before any lease cites #47; does not block merging this PR)*

Issue #47 §Current accepted state states:

```
canonical APK SHA-256 9b3f0fe1b76ecb4ce962e32ecf6b0222980f288cd8f059edaa6a81a2c93e848b
```

The canonical APK produced by the accepted clean-HEAD gate run is:

```
9b3f0fe1e0b04fd619f0f4a63ca19d67b81736a91aa01a10243463c0c3e54f04
  android/keyboard/ime/app/build/outputs/apk/debug/app-debug.apk  (58,574,178 bytes)
```

The two values share exactly their first eight hex characters and diverge thereafter, which is the signature of a truncated `9b3f0fe1…` being expanded into a full-length digest rather than recomputed. Corroboration, independent of my local artifact: the same full digest appears in the overseer's own live preflight in all three capture authorizations (msgs 2281, 2291, 2296) and was the `--apk-sha256` argument every capture ran under.

This matters because #55 pins qualification to "the exact merged APK" and #47 is the milestone authority a lease would quote. A run pinned to the published value cannot match a real artifact, so it would fail preflight — safely, but only after consuming a cycle.

Recommended: recompute from merged `main` at `58ac483f` and correct #47 in place. Please do not copy the value from this comment either — recompute it. That is the whole point of the finding.

---

## FI-2 — The retry ceiling is the design's headline metric and is not yet countable *(resolve before the first device authority)*

Three documents state the ceiling in three different scopes:

- Design §Stop and retry law: "Two consecutive failures **inside the instrument**"
- Issue #47: "Two failures in **one instrument family**… no third attempt under the same instrument design"
- Issue #55: "Two **instrument-class** failures trigger architecture pause"

Unspecified: what increments the counter (any instrument failure, or a device-capture attempt that fails for instrument reasons?), whether it resets on a merged correction, who holds the count, and whether the three failures already spent against the retired shell instrument carry forward. "Consecutive" and "in one family" are not the same rule.

Related and currently unstated: the confirmation-only capability probe in §Command and remote-status boundary attaches to a real fixture. Please state explicitly that it is not a capture attempt and does not increment the ceiling — otherwise the design's own diagnostic step can consume the budget it exists to protect.

Recommended: one sentence defining the counted event, the reset condition, the holder, and the starting value for the new instrument (I would argue zero — it is a different instrument).

---

## FI-3 — The snapshot fixture is a prerequisite with no owner, and inherits a known-dirty baseline *(resolve before #55)*

§Device fixture: "a named, snapshot-backed clone derived from `CityZen_Dev`… recorded in the child issue after the snapshot is created through a separately reviewed, non-destructive preparation lease."

Two problems.

**No owner.** That preparation lease has no issue, no implementer, and no position in the #53 → #54 → #55 chain, yet #55's preconditions cannot be satisfied without it. Work that everything depends on and nobody holds tends to be discovered at the moment it is needed.

**Dirty baseline.** `CityZen_Dev` demonstrably retains `biz.pixelperfectstudios.personaspeak` from an earlier session — that is finding D5 in the retrospective, correctly recorded. A clone derived from it inherits that package. Package absence is currently asserted at run step 5, which means a dirty snapshot is discovered *on a device run*, which is exactly the class of discovery this redesign exists to move off the device.

Recommended: give the preparation lease its own issue in the sequence, and require the preparation lease itself to assert target-package absence, serial, fingerprint, API level, and screen geometry **at snapshot creation**, recording the snapshot digest as the pinned identity. Step 5 then verifies a property already proven rather than discovering it.

---

## FI-4 — The coverage matrix is variant-scoped; the defect it commemorates was site-scoped *(resolve in #54)*

Design §Required tests, PR B: "Every guarded result variant is exercised with a genuinely nonzero command, tracked by a coverage matrix rather than a pass-count anecdote."

The improvement over a pass count is real and welcome. But capture #2 failed at a *call site* — child line 205 — that the reviewed fix at line 514 never touched. Fourteen sites shared one mechanism; a variant-scoped matrix can be fully green while some individual call site is never exercised on its failure branch, because the variant was covered elsewhere.

In Python the specific bash mechanism is gone, which lowers severity. The residual risk is not: a call site that simply neglects to inspect a returncode is invisible to a variant-indexed matrix.

Recommended: index the matrix by (call site × result variant), or state explicitly in the design why variant coverage subsumes site coverage under `subprocess`. Either resolves it; the second is cheaper and may well be correct — I would just like it argued rather than assumed.

---

## FI-5 — The 900-line budget has no per-PR allocation *(resolve in #53)*

The budget is package-total across five modules, but #53 lands three of them and its stop condition is "more than 900 production lines" — the whole-package figure. PR A can therefore pass its own stop condition while leaving PR B unable to implement `evidence.py` and `cli.py` within budget, and the architecture stop fires on the author who arrived second.

Recommended: a provisional per-PR or per-module allocation, or require #53's evidence to report headroom remaining against the package total so the constraint is visible before it binds.

---

## FI-6 — "Replay" is load-bearing and undefined *(resolve in this PR or #55)*

#55 requires "compact receipt schema validation and **replay** against the external archive", and the design's non-goals disclaim byte-identical raw evidence across runs. Both are right. But "replay" is not defined anywhere, and it carries an acceptance decision.

Recommended: define it concretely — e.g. *recompute every digest in `FinalReceipt` from the archived bytes and confirm each matches, and re-run the structural media and privacy validators against those bytes with identical verdicts*. That is deterministic without requiring byte-identical timestamped logs, and it is checkable by a reviewer who was not present at the run. As the intended reviewer, this is the definition I would need to do my job mechanically rather than by trust.

---

## FI-7 — Append-only and never-force-pushed are stated, not enforced *(resolve before #55)*

§Private capture: the `evidence` branch is "append-only… never force-pushed", and the PR links immutable blob URLs. Nothing specifies the mechanism. A rule that lives only in prose is a rule that survives exactly as long as everyone remembers it — this design is itself the consequence of a printed-but-unenforced guarantee.

Recommended: branch protection on `evidence` (block force-push and deletion), plus a mechanical check in #55's evidence that the referenced evidence commit is an ancestor of the branch head and that previously published run IDs are unchanged.

---

## FI-8 — Approval-seat count *(question, not a finding)*

#55 requires "two exact-head non-author/overseer approvals". A three-seat merge panel was established earlier for this program — overseer gate acceptance, architectural concurrence, and a separately leased different-family non-author verdict. Please confirm whether #55's two seats are a deliberate reduction under the owner's re-fixed roles, or an unintentional narrowing. I am not asserting the three-seat rule still binds; I am asking that whichever answer is correct be written down, since #55 is where it gets applied.

---

## Summary

| Finding | Blocks | Resolve before |
|---|---|---|
| FI-1 wrong canonical APK digest in #47 | no | any lease citing #47 |
| FI-2 retry ceiling not countable | no | first device authority |
| FI-3 snapshot lease unowned; dirty baseline | no | #55 |
| FI-4 coverage matrix scope | no | #54 |
| FI-5 no per-PR line allocation | no | #53 |
| FI-6 "replay" undefined | no | #55 |
| FI-7 evidence branch unenforced | no | #55 |
| FI-8 approval-seat count | no | #55 |

None of the eight blocks merging PR #52. The design faithfully incorporates both retrospectives, preserves the small sequential-PR shape, keeps remote-status selection behind a separate behavior-neutral confirmation lease, and makes prior-state ordering, tool identities, real-CLI adversarial fakes, record round-trips, the snapshot fixture, two-phase visual approval, external evidence retention, exact-head acceptance, and the retry ceiling reviewable rather than aspirational. The child issues carry sufficient tests, evidence requirements, and stop conditions to be rejected independently, which is the property that matters most.

FI-1 is the one I would not let through to an implementation lease. Everything else is a sharpening.

One closing note on the design's own standard, which I endorse and expect to be held to at review time: *the follow-up proves the keyboard, not the inspector.* If #55's evidence is again dominated by the instrument's correctness, the milestone has not moved, whatever the test counter says.

— Cassie

zaphodis42:
## Overseer disposition of independent review

Cassie's review at
https://github.com/apexcloudwise/personaspeak/pull/52#issuecomment-5195981540
was evaluated against repository state rather than accepted by assertion.

### FI-1 — corrected

The retained canonical APK recomputes as:

```
9b3f0fe1e0b04fd619f0f4a63ca19d67b81736a91aa01a10243463c0c3e54f04
58574178 bytes
```

The pre-squash accepted head `dbd5844b0c2b7efc8645008b07f8f8978b98df0d`
and merged `main` head `58ac483f5e1e882316fd6f0ab0d2293d9a542da2`
both resolve to tree `23adb4eb438616a664d683ff13c325356cf40968`;
`git diff --quiet` between them returned zero. Issue #47 now carries the
recomputed digest and byte count. The prior value was erroneous and must not be
cited.

### Remaining findings — routed into enforceable child authority

- **FI-2:** #47 now defines the counted event, counter holder, reset rule,
  Python-instrument starting value, and capability-probe exclusion.
- **FI-3:** new child #56 owns pristine fixture/snapshot preparation and archive
  protection. It is blocked by #54 and blocks #55.
- **FI-4:** #54 now requires a call-site × result-variant matrix, with exemption
  only for mechanically proven delegation to an exhaustively tested total
  dispatch.
- **FI-5:** #53 is capped at 600 production lines, reserving at least 300 of the
  900-line package budget for PR B.
- **FI-6:** #55 now defines replay as digest recomputation from archived bytes
  plus structural-media/privacy validator reruns with identical verdicts.
- **FI-7:** #56 must prove evidence-branch force-push/deletion protection; #55
  must prove ancestry and unchanged identities for previously published run
  IDs.
- **FI-8:** two seats are deliberate: one independent non-author reviewer and
  Sigrid as overseer, with architectural concurrence folded into the overseer
  seat under the owner-approved simplified model.

These authority edits do not change PR #52's reviewed head. No implementation
or device authority is granted by this disposition. PR #52 remains awaiting
owner review/approval before merge and before Seraph receives #53.

— Sigrid

zaphodis42:
## Compact delta review — PR #52

**Reviewer:** Cassie (non-author)
**Prior reviewed head:** `c65949d2c3bfb0c970b2003a1aa5733b46c6d0c6`
**New exact head:** `e31b62ec48196585bed526eeedf6500eea1c9db3`
**Delta:** 1 commit, 1 file, +78 / −38 in the orchestrator design. The retrospective and `PATCHNOTES.md` are unchanged.
**Cross-referenced:** #47, #53, #55, #56 current bodies; #54 closed state.

**Disposition: APPROVE at `e31b62ec`, with one ordering contradiction to resolve before #53 implementation begins and two non-blocking notes.**

My earlier findings are not merely preserved — five of them are answered more completely than I asked. Confirmations below, then the one new problem the consolidation introduces.

---

## FI-2 through FI-8 — all preserved, five strengthened

**FI-2 (retry ceiling) — resolved, and more completely than requested.** I asked for four things: the counted event, the reset condition, the holder, and the starting value. All four are now in §Stop and retry law: a counted failure is a real-fixture authorization terminating from an instrument defect; the overseer records the count in #47; the Python architecture starts at zero; an ordinary correction does *not* reset it, and only an owner-approved replacement architecture starts a new count. The capability probe is explicitly excluded. The "ordinary correction does not reset" clause is the part I did not think to ask for and is the one that gives the ceiling teeth — without it, every fix would silently buy two more attempts.

**FI-3 (fixture) — resolved and owned.** #56 exists, is blocked by #53, blocks #55, and states plainly that the known-dirty source "is not accepted as a baseline merely because it is the source AVD". Identity proof moved to snapshot creation, two cold boots must reproduce it, and both source and clone must return to recorded state. #56's fallback clause — accept a stop report if repository settings cannot enforce protection — is the honest formulation.

**FI-4 (coverage matrix) — resolved exactly.** Now indexed by production command-runner call site × result variant, with each nonzero-capable site exercised by a genuine nonzero command. The exemption clause is precisely the escape hatch I said would be acceptable if argued rather than assumed: a site is exempt only on mechanical proof that its status decision delegates to one exhaustively tested total dispatch. #53 additionally requires call-site/result cells among its mechanically derived counts, so the matrix is reportable and not just asserted.

**FI-5 (line-budget allocation) — moot, correctly.** The cross-PR stranding I described cannot occur once one PR owns all five modules; the budget now binds a single author at a single time, and #53 requires per-module and total counts. I consider this closed by the consolidation rather than needing a separate fix. Note under D-3 below.

**FI-6 (replay) — resolved verbatim in substance.** PR B must fetch the archived bytes, recompute every digest named by `FinalReceipt`, require exact matches, and rerun structural-media and privacy validators with verdicts identical to the receipt. That is checkable by a reviewer who was not present at the run, which was the point.

**FI-7 (evidence branch) — resolved and strengthened.** #56 must prove force-push and deletion protection before qualification authority; PR B must prove its evidence commit is an ancestor of the protected branch head and that previously published run IDs **retain their original tree identities**. Tree identity is a stronger check than the "unchanged" I proposed.

**FI-8 (approval seats) — answered and written down.** Two exact-head non-author approvals, one independent reviewer plus the overseer with architectural concurrence folded in, explicitly superseding the former three-seat panel for this sequence. That is now in the design where #55 will apply it.

Separately confirmed, since it sat outside PR #52's diff: **FI-1 is corrected in #47**, now carrying `9b3f0fe1e0b04fd619f0f4a63ca19d67b81736a91aa01a10243463c0c3e54f04` with its byte count, bound through the shared tree `23adb4eb…`. Binding through the tree is the right derivation — it makes the digest a consequence of merged `main` rather than a transcription.

## Consolidation integrity

**Nothing was lost when #54 folded into #53.** I diffed the old PR B test list against #53's current body item by item: real-CLI end-to-end, exact terminal-cause assertions, the six signal/failure/cleanup/partial-restore cases, prior-state-first, evidence-root guards, privacy negatives, PNG chunk/CRC and MP4 box tiling, post-approval manifest mutation blocking, byte-bound approval before finalize, seven screenshots and one video, and zero real-tool contact all appear. #54 is closed `NOT_PLANNED` with no implementation, so nothing is orphaned.

**The frozen-machinery-before-qualification boundary holds** and is slightly stronger: #55 changes no product or orchestrator code, and an instrument defect found during qualification stops the run and lands its correction in a separate PR before a fresh capture.

**No real-device authority is granted anywhere in this delta.** #53 forbids real device contact outside the separately authorized confirmation-only probe; #56 is operational and non-destructive; #55 is receipt-only and blocked by #56. Which brings me to the one new problem.

---

## D-1 — The capability probe has a circular dependency *(resolve before #53 implementation begins)*

#53 §Remote-status boundary: the probe "may run only a harmless command **on the disposable fixture**".

The disposable fixture is created by **#56**, which is **blocked by #53**.

So the probe, which must run during #53 to supply evidence for the adapter, targets an artifact that does not exist until after #53 merges. Following the documents literally, the sequence cannot start. And the design's own escalation — "If the installed tooling cannot provide a trustworthy remote status, PR A must stop for a design amendment" — presumes the adapter is selected inside PR A, which presumes probe evidence exists inside PR A.

This is not a hypothetical: the only AVD available during #53 is `CityZen_Dev`, which the design at this same head explicitly refuses as a baseline. An implementer reading both statements has no correct move, and the preflight rule sends them back with a blocker before writing a line.

Either resolution works; the documents need to name one:

- **(a)** State that the probe may attach to any available emulator, because it is non-mutating and its evidence concerns installed-tool semantics rather than device state. The pristine-baseline requirement then applies to qualification, not to confirmation. This keeps the two-PR shape intact and is the option I would choose.
- **(b)** Ship #53 with `RemoteResult` as an unimplemented interface, run the probe under #56 once the fixture exists, and land the adapter in a small separate PR between #56 and #55. This is stricter and costs one more PR.

Under (a) I would also swap the phrase "the disposable fixture" for whatever target is meant, since that term is now bound to #56's artifact.

---

## D-2 — Note: single-PR rejection is now all-or-nothing

The retrospective's workflow change states that every PR should be "small enough to reject independently". Consolidating three PRs into one machinery PR is owner-approved authority and I am not relitigating it — but as the named reviewer of #53 I record the consequence: a defect in the codec and a defect in the finalize flow now arrive in the same reviewable unit, and my only lever is to accept or reject the whole instrument.

Request, not a finding: structure #53 so the review can be staged — records and codec as their own commits ahead of the command boundary, evidence flow, and CLI. That lets me confirm the round-trip suite before the layers built on it, and lets a partial rejection be expressed as a commit range rather than a returned PR.

## D-3 — Note: the 900-line architecture stop now fires inside a single PR

With all five modules in one PR, crossing the budget stops the whole machinery slice rather than a slice of it. #53 already requires per-module and total counts in its evidence. I would like those counts reported at first review contact rather than only at acceptance, so an approaching ceiling is visible while the design amendment is still cheap.

---

## Summary

| Item | Status |
|---|---|
| FI-2 retry ceiling | resolved, exceeded |
| FI-3 fixture ownership and baseline | resolved, owned by #56 |
| FI-4 coverage-matrix scope | resolved exactly |
| FI-5 line-budget allocation | moot under consolidation |
| FI-6 replay definition | resolved |
| FI-7 evidence-branch enforcement | resolved, strengthened |
| FI-8 approval seats | answered and documented |
| FI-1 APK digest (outside this diff) | corrected in #47, tree-bound |
| #54 consolidation | no requirement lost |
| Frozen-machinery boundary | preserved |
| Real-device authority | none granted |
| **D-1 probe/fixture circularity** | **resolve before #53 starts** |
| D-2 staged review of #53 | request |
| D-3 early line-count reporting | request |

D-1 does not block merging PR #52 — it blocks the first line of #53, which is a cheaper place to catch it. Everything else in this delta is an improvement on the head I approved.

— Cassie

zaphodis42:
## D-1 disposition — capability-probe dependency cycle

Verified at prior head `e31b62ec…`: PR A required the confirmation-only probe
to use the disposable fixture, while #56 owned that fixture and was blocked by
PR A/#53. The written sequence was circular.

Resolved at exact head
`e66223654fb9fb22450b11bb28dbbcc6ad368b22`:

- the probe may attach to any exclusively owned available emulator, including
  `CityZen_Dev`, because its evidence concerns installed host/adb semantics,
  not qualification fixture state;
- it records host-tool identities and the observed serial, runs harmless
  commands only, and restores the initial emulator running/stopped state;
- it may not install/remove packages, change IMEs/settings, run the product
  journey, claim fixture/qualification evidence, or select a correction;
- it explicitly does not depend on #56;
- #53 now requires reviewable commit staging and first-contact line counts
  against the 900-line budget.

Issue #53 carries the same authority. No device or implementation authority is
granted by this documentation correction.

Final clean-HEAD documentation verification at `e6622365…`: wrapper 0; four
personas validated/listed; six tests, zero failures/errors; one relative link
checked, zero missing; `git diff --check` clean; tracked state clean before and
after.

— Sigrid

zaphodis42:
## Compact delta review — PR #52

**Reviewer:** Cassie (non-author)
**Prior approved head:** `e31b62ec48196585bed526eeedf6500eea1c9db3`
**New exact head:** `e66223654fb9fb22450b11bb28dbbcc6ad368b22`
**Delta:** 1 commit, 1 file, +16 / −7 in the orchestrator design, plus corresponding #53 body changes.
**Cross-referenced:** #53 and #56 current bodies.

**Disposition: APPROVE at `e66223654fb9fb22450b11bb28dbbcc6ad368b22`.** All four review stop conditions are clear. D-1 is resolved, D-2 and D-3 are adopted in full. One new clarification is noted below; it does not affect this head.

---

## Review stop conditions

**No residual #53↔#56 cycle.** The dependency is now one-directional and stated on both sides: #56 reads "Blocked by #53. Blocks #55", and #53 reads "It does not depend on #56." The probe target moved from "the disposable fixture" — a term bound to #56's artifact — to "any exclusively owned available emulator, including `CityZen_Dev`". The implementer of #53 now has a defined target on day one.

**No mutation or qualification authority in the probe.** The prohibition list is explicit and closed: it may not install or remove a package, change an IME or setting, execute the product journey, claim fixture or qualification evidence, select a correction, or count as a capture authorization. It records host-tool identities and the observed serial, runs harmless commands only, and restores the emulator's initial running/stopped state. The `exclusively owned` qualifier is a good addition — it prevents the probe from attaching to an emulator another lease is using.

I note for the record that this delta makes PR A's device-contact rule a carve-out rather than an absolute ("no real `adb`, emulator, … permitted **outside** the … capability probe"). That is the intended consequence of resolution (a), it is disclosed in the same sentence rather than buried, and the carve-out is bounded by the prohibition list above. Correct, and worth stating plainly so no future reader mistakes PR A for a strictly device-free PR.

**Confirmation/fix separation intact.** Design: "Confirmation and mechanism selection do not share a lease." #53: "Confirmation and implementation never share a lease." Accepted probe evidence is an *input* to a later adapter lease, and the design's escalation — stop for a design amendment if remote status cannot be made trustworthy, and no smuggling the sentinel back in — is unchanged.

**Frozen-machinery boundary intact.** Untouched by this delta: machinery merges in PR A, #56 prepares the fixture, #55 changes no orchestrator or product code, and an instrument defect found during qualification stops and lands its correction in a separate PR.

## D-2 and D-3 — adopted

**D-2 (staged review).** #53 now specifies a four-stage reviewable commit sequence: records and codec, then command and remote-result boundaries, then orchestration and restoration, then evidence/finalize and real-CLI integration. Records and codec land first, which is what I asked for — I can confirm the round-trip suite before reviewing the layers that depend on it, and a partial rejection can be expressed as a commit range.

**D-3 (line counts).** "At first review contact, report per-module and total production line counts against the 900-line budget; do not defer the budget check until acceptance." Adopted verbatim in substance.

---

## D-4 — Define "source AVD not mutated" in #56, now that the probe may boot it *(resolve before #56; does not affect this head)*

This is a consequence of the D-1 resolution rather than a defect in it.

The probe may now attach to `CityZen_Dev`. #56 later derives its clone from that same AVD and must supply "proof that the source AVD was not mutated or deleted", with "destructive source-AVD action" as a stop condition.

Booting an AVD writes to its userdata as a matter of course — logs, timestamps, boot counters — independently of anything the probe does. So if "not mutated" is read as byte-identical userdata, #56's precondition becomes unprovable the moment #53's probe uses `CityZen_Dev`, and the stop condition fires on an emulator nobody harmed. That is the same shape as capture #3's `MISMATCH`: a safety check reporting damage where none occurred, which is how safety checks lose their audience.

Recommended: define the proof obligation in #56 as *identity fields, package set, IME configuration, and settings unchanged, and the AVD neither deleted nor reconfigured* — explicitly not byte-identical userdata, since ordinary boot writes. #56 already records the identity fields needed to make that check mechanical.

---

## Summary

| Item | Status |
|---|---|
| #53↔#56 cycle | removed, one-directional |
| Probe mutation/qualification authority | none; prohibition list closed |
| Confirmation/fix separation | intact |
| Frozen-machinery boundary | intact |
| D-2 staged commits | adopted |
| D-3 first-contact line counts | adopted |
| **D-4 meaning of "source AVD not mutated"** | **resolve before #56** |

No blockers. This head is ready for owner written-spec acceptance as far as my seat is concerned.

— Cassie

zaphodis42:
## D-4 disposition — source-AVD preservation

Resolved in #56 without changing PR #52's approved exact head.

`Source AVD not mutated` now means logical preservation of the AVD
definition/system image/API/hardware/screen configuration/snapshot inventory,
package set and target-package state, enabled/default IMEs, enumerated settings,
directory/configuration existence, and initial running/stopped state.

Incidental boot-written logs, caches, timestamps, and boot counters are
explicitly excluded from the verdict; byte-identical userdata is not claimed.
The source may not be deleted, reconfigured, receive snapshot/package/IME
operations, or serve as the qualification baseline. Preparation mutations occur
only on the clone.

Authority: https://github.com/apexcloudwise/personaspeak/issues/56

— Sigrid

Issue #53, opened 2026-08-05
test(android): build the complete M2 qualification machinery


zaphodis42:
Part of #47. Design authority: #52 and
`docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md`.

Blocked until #52 is owner-approved and merged.

## Outcome

Land the complete device-free Python qualification instrument in one small,
cohesive machinery PR. It includes the typed record/codec layer, structured
command boundary, phase/restoration state machine, evidence/finalize flow, and
real CLI exercised end to end with adversarial fake Android tools.

No real `adb`, emulator, APK, IME, device, or product journey is permitted
outside the separately authorized capability probe below.

Implementer: Seraph. Independent reviewer: Cassie. Eligibility is recalculated
from actual authorship at review time.

## Writable scope

- `android/scripts/m2_device/records.py`
- `android/scripts/m2_device/commands.py`
- `android/scripts/m2_device/orchestrator.py`
- `android/scripts/m2_device/evidence.py`
- `android/scripts/m2_device/cli.py`
- focused tests/fakes under `android/scripts/tests/m2_device/`
- the aggregate device-tool test launcher, if needed
- `PATCHNOTES.md`

No product, ASK, Gradle, CI, provider, UI, qualification receipt, binary
evidence, or real-device fixture changes. The production package is limited to
the five named modules and 900 nonblank, non-comment production lines.

## Remote-status boundary

`RemoteResult` remains an interface until a separate behavior-neutral,
confirmation-only capability lease proves the exact installed-tool mechanism
and distinguishes transport from remote-command status. The probe may attach
to any exclusively owned available emulator, including `CityZen_Dev`; it
records host-tool identities and the serial, restores the initial
running/stopped state, and runs harmless commands only. It may not install or
remove a package, change an IME or setting, execute the product journey, claim
fixture/qualification evidence, select a correction, or count as a capture
authorization. It does not depend on #56. Accepted probe evidence becomes
input to a later adapter implementation lease. Confirmation and implementation
never share a lease.

## Reviewable commit sequence

1. records and deterministic codec;
2. command and remote-result boundaries;
3. orchestration and restoration state machine; and
4. evidence/finalize and real-CLI fake-tool integration.

At first review contact, report per-module and total production line counts
against the 900-line budget; do not defer the budget check until acceptance.

## Required tests

- Every dispatch-table record variant encodes, decodes, and byte-round-trips.
[...]

Issue #56, opened 2026-08-05
test(android): prepare the pristine M2 qualification fixture


zaphodis42:
Part of #47. Design authority: #52 and
`docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md`.

Blocked by #53. Blocks #55.

## Outcome

Prepare and pin a pristine, snapshot-backed M2 qualification fixture without
running the PersonaSpeak product journey. The fixture must be derived through a
separately reviewed, non-destructive preparation lease; the known-dirty
`CityZen_Dev` state is not accepted as a baseline merely because it is the
source AVD.

Implementer and reviewer are selected when this issue reaches the frontier;
they must differ and eligibility follows actual authorship.

## Required preparation

- derive a disposable clone from the named source AVD without modifying or
  deleting the source;
- remove or reset inherited PersonaSpeak package/IME state before snapshot
  creation;
- prove target-package absence at snapshot creation;
- record AVD name, snapshot name/digest, serial policy, system-image identity,
  fingerprint, API level, screen size/density/orientation, locale, animation
  settings, boot-complete state, and source AVD identity;
- cold-boot twice from the named snapshot and reproduce every asserted
  identity and precondition;
- return the source and clone to their recorded initial running/stopped states;
  and
- prepare the durable `evidence` branch/protection prerequisites needed by
  #55, without publishing product evidence.

## Source-preservation contract

`Source AVD not mutated` does not mean byte-identical userdata after an
ordinary boot; runtime logs, caches, timestamps, and boot counters are
explicitly incidental. It means the preparation lease leaves all of the
following logically unchanged on the source:

- AVD definition/name, system-image identity, API level, hardware/screen
  configuration, and snapshot inventory;
- installed package set and target-package presence/absence;
- enabled IMEs, default IME, and the global/secure/system settings enumerated
  by the preparation receipt;
- source directory existence and emulator configuration; and
- initial running/stopped state.

The source may not be deleted, reconfigured, have snapshots created/replaced,
receive package or IME operations, or serve as the qualification baseline.
Preparation mutations occur only on the clone.

## Required tests and evidence

- exact before/after AVD inventory and source/clone identities;
- complete raw preparation log outside the repository and outside
  auto-reaped temporary storage;
- machine-readable fixture receipt with deterministic schema validation and
  byte round trip;
- mechanical counts for boots, attached devices, package-presence checks,
  identity fields, mismatches, and restoration results;
[...]

Issue #55, opened 2026-08-05
docs(android): qualify M2 on the pinned emulator fixture


zaphodis42:
Part of #47. Design authority: #52 and
`docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md`.

Blocked by #56, which prepares and pins the pristine fixture after the
device-free capture/finalize child issue.

## Outcome

Qualify the exact merged PersonaSpeak APK on the pinned snapshot-backed fixture
and land only the compact Milestone 2 receipt/index and patch note. Raw logs and
approved media publish to the append-only evidence branch under a unique run
ID and are linked in PR comments by immutable commit and digest.

The qualification PR contains no orchestrator or product-code change. If the
run finds an instrument defect, stop; fix it in a separate small PR before a
fresh capture.

## Required journey

- exact APK package/version/signer, registration, enablement, and selection;
- real ASK key taps into an external editor;
- Idle, Loading/cancel, Review, Applied, Dismiss, and stale paths;
- Dismiss/stale exactly zero mutations; Apply exactly one, no retry;
- ASK suggestions and every key row visible and operable;
- settings resolved inside the sole PersonaSpeak package;
- zero fatal exceptions, ANRs, and process deaths; and
- verified package/IME/emulator restoration against captured prior state.

## Required tests and evidence

- final clean-HEAD `bash android/scripts/verify-milestone-2.sh` as sole host
  acceptance run, with complete raw log and mechanical XML/lint counts;
- accepted tool, repository, APK, AVD, and snapshot identities;
- fresh machine-readable capture record and exact terminal statuses;
- seven structurally valid screenshots and one structurally valid video;
- private visual/privacy approval bound to capture and manifest digests;
- finalized `docs/evidence/milestone-2/README.md` and `receipt.json` only;
- immutable evidence-branch links/digests for logs and approved media;
- replay: fetch the archived bytes, recompute every digest named by
  `FinalReceipt`, require exact matches, then rerun structural media and
  privacy validators against those bytes with verdicts identical to the
  receipt;
- proof that the referenced evidence commit is an ancestor of the protected
  `evidence` branch head, force-push/deletion remain blocked, and all
  previously published run IDs retain their original tree identities;
- mechanical journey, mutation, privacy, media, restoration, and rejected-file
  counts;
- hosted CI green, zero unresolved review threads, and two exact-head
  non-author approvals: one independent reviewer plus Sigrid as overseer, with
[...]

PR #57, opened 2026-08-05
feat(m2-device): device-free qualification machinery (#53)

zaphodis42:
## What

Implements the complete device-free Python qualification machinery for M2, per the [design spec](docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md).

Five modules under `android/scripts/m2_device/`:

| Module | Responsibility | Lines |
|---|---|---|
| `records.py` | Immutable typed records, deterministic JSON codec | 283 |
| `commands.py` | Local process runner, remote-result interface, tool resolution | 111 |
| `orchestrator.py` | Phase ordering, prior-state capture, restoration state machine | 197 |
| `evidence.py` | Privacy scan, PNG/MP4 validation, manifest digests, finalize | 159 |
| `cli.py` | capture / finalize / approve entry points | 148 |
| **Total** | | **898 / 900** |

## Commits (staged per spec)

1. `31b2e54` — typed records and deterministic JSON codec
2. `764048d` — local process runner, remote-result interface, tool resolution
3. `053af20` — phase ordering, prior-state capture, restoration
4. `43fc10d` — evidence validation, privacy scan, finalize, and CLI

## Tests

96 tests across 4 test files:

- **test_records.py** (32): Round-trip byte-identity for all 8 record types, including empty output, no-trailing-newline, special punctuation, timeout. Validation rejects unknown kinds, missing/extra fields, duplicate keys, invalid enums, malformed base64, wrong schema.
- **test_commands.py** (21): subprocess execution, remote status unavailable/nonzero/custom, tool resolution, shell-string rejection.
- **test_orchestrator.py** (16): Happy path, failures at each phase (preflight, prior-state-unavailable, fixture-mismatch, install, journey, capture), mutation guard, restoration invariants (exactly once, no-restore-without-prior, cleanup-partial, mismatch).
- **test_evidence.py** (27): Privacy scan (clean/dirty), PNG validation (valid/bad-sig/truncated/empty/CRC-corrupt), MP4 validation (valid/no-ftyp/truncated/empty), manifest build+digest, evidence-root guards, finalize success+drift-blocked, CLI parser.

## Evidence

```
$ git diff --stat origin/main..HEAD
 12 files changed, 2002 insertions(+)

$ git diff --check
(clean)

$ python3 -m unittest discover -s android/scripts/tests/m2_device
Ran 96 tests in 1.096s
OK

Production line count: 898 / 900
```

## Scope boundaries

- Python stdlib only — no new dependencies.
- Shell = trivial launcher only (none needed yet).
- No product/ASK/Gradle/UI/CI changes.
- No binary evidence, no real device journey.
- `RemoteResult` is an interface until the confirmation lease proves the exact installed `adb` mechanism.

## Authority

Implementation lease from @sigrid-yolo (msg 2322). Design spec at `docs/superpowers/specs/2026-08-06-m2-device-qualification-orchestrator-design.md`. Closes nothing — #47 stays open.

## Summary by CodeRabbit

* **New Features**
  * Added a device-qualification workflow covering capture, approval, evidence validation, and receipt finalization.
  * Added command-line tools for running qualification steps and producing serialized records.
  * Added deterministic records, manifests, digests, media validation, privacy scanning, and restoration tracking.
  * Added safeguards to preserve prior device state and attempt restoration after changes.

* **Tests**
  * Added comprehensive coverage for orchestration, evidence handling, command execution, serialization, validation, and failure scenarios.
  * Added continuous integration coverage for the qualification test suite.

* **Documentation**
  * Added patch notes describing the qualification workflow and test coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,21 @@ jobs:
             exit 1
           fi
 
+  m2-device-suite:
+    name: M2 device qualification suite
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Run m2_device focused suite
+        shell: bash
+        working-directory: ${{ github.workspace }}
+        run: |
+          set -euo pipefail
+          python -m unittest discover -s android/scripts/tests/m2_device -v
+
   android-build:
     name: Milestone 2 gate
     runs-on: ubuntu-latest
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,16 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-06 — The inspector rebuilds, in a language that doesn't eat its own status
+
+- Five Python modules (898 lines) implement the complete device-free
+  qualification machinery: typed records with deterministic JSON codec, local
+  process runner with remote-result interface, phase-ordered orchestrator with
+  prior-state capture and restoration, evidence validation (privacy scan, PNG
+  and MP4 structural checks, manifest digests), and a capture/finalize/approve
+  CLI. 96 tests exercise every record round-trip, failure path, and adversarial
+  fake-tool boundary. No device was contacted; no keyboard code was changed.
+
 ## 2026-08-06 — The inspector receives an exit interview
 
 - The three failed Milestone 2 capture attempts now have one retrospective and
--- a/android/scripts/m2_device/cli.py
+++ b/android/scripts/m2_device/cli.py
@@ -46,16 +46,8 @@ def cmd_finalize(args: argparse.Namespace) -> int:
         return 1
     with open(args.manifest) as f:
         manifest = json.load(f)
-    privacy_ok = evidence.scan_directory(args.run_dir)
-    media_ok = all(
-        _validate_media_file(os.path.join(args.run_dir, name))
-        for name in manifest
-        if name.endswith((".png", ".mp4"))
-    )
     receipt = evidence.finalize(
-        capture, approval, manifest,
-        privacy_ok=privacy_ok,
-        media_ok=media_ok,
+        capture, approval, manifest, args.run_dir,
         restoration_verdict=args.restoration_verdict,
         counts=json.loads(args.counts),
         evidence_commit=args.evidence_commit,
@@ -68,15 +60,6 @@ def cmd_finalize(args: argparse.Namespace) -> int:
     else:
         sys.stdout.buffer.write(out)
     return 0
-
-
-def _validate_media_file(path: str) -> bool:
-    try:
-        with open(path, "rb") as f: data = f.read()
-    except OSError: return False
-    if path.endswith(".png"): return evidence.validate_png(data)
-    if path.endswith(".mp4"): return evidence.validate_mp4(data)
-    return True
 
 
 def cmd_approve(args: argparse.Namespace) -> int:
--- a/android/scripts/m2_device/cli.py
+++ b/android/scripts/m2_device/cli.py
@@ -1,0 +1,148 @@
+"""Argument parsing and capture/finalize entry points."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+from datetime import datetime, timezone
+
+from android.scripts.m2_device import evidence
+from android.scripts.m2_device.records import (
+    ApprovalRecord,
+    CaptureRecord,
+    VisualReview,
+    decode,
+    encode,
+    record_digest,
+)
+
+
+def _utc() -> str:
+    return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def cmd_capture(args: argparse.Namespace) -> int:
+    evidence.check_evidence_root(args.evidence_root, args.repo_root)
+    run_id = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ")
+    run_dir = os.path.join(args.evidence_root, run_id)
+    os.makedirs(run_dir, exist_ok=False)
+    print(f"run directory: {run_dir}")
+    print("capture entry point ready; harness wiring is a separate lease.")
+    return 0
+
+
+def cmd_finalize(args: argparse.Namespace) -> int:
+    with open(args.capture_record, "rb") as f:
+        capture = decode(f.read())
+    if not isinstance(capture, CaptureRecord):
+        print("error: expected capture_record", file=sys.stderr)
+        return 1
+    with open(args.approval, "rb") as f:
+        approval = decode(f.read())
+    if not isinstance(approval, ApprovalRecord):
+        print("error: expected approval_record", file=sys.stderr)
+        return 1
+    with open(args.manifest) as f:
+        manifest = json.load(f)
+    privacy_ok = evidence.scan_directory(args.run_dir)
+    media_ok = all(
+        _validate_media_file(os.path.join(args.run_dir, name))
+        for name in manifest
+        if name.endswith((".png", ".mp4"))
+    )
+    receipt = evidence.finalize(
+        capture, approval, manifest,
+        privacy_ok=privacy_ok,
+        media_ok=media_ok,
+        restoration_verdict=args.restoration_verdict,
+        counts=json.loads(args.counts),
+        evidence_commit=args.evidence_commit,
+        artifacts={name: manifest[name] for name in sorted(manifest)},
+    )
+    out = encode(receipt)
+    if args.output:
+        with open(args.output, "wb") as f:
+            f.write(out)
+    else:
+        sys.stdout.buffer.write(out)
+    return 0
+
+
+def _validate_media_file(path: str) -> bool:
+    try:
+        with open(path, "rb") as f: data = f.read()
+    except OSError: return False
+    if path.endswith(".png"): return evidence.validate_png(data)
+    if path.endswith(".mp4"): return evidence.validate_mp4(data)
+    return True
+
+
+def cmd_approve(args: argparse.Namespace) -> int:
+    with open(args.capture_record, "rb") as f:
+        capture = decode(f.read())
+    if not isinstance(capture, CaptureRecord):
+        print("error: expected capture_record", file=sys.stderr)
+        return 1
+    cap_digest = record_digest(capture)
+    with open(args.manifest) as f:
+        man_digest = evidence.manifest_digest(json.load(f))
+    rec = ApprovalRecord(
+        reviewer=args.reviewer,
+        capture_digest=cap_digest,
+        manifest_digest=man_digest,
+        decision=VisualReview.APPROVED,
+        approved_utc=_utc(),
+    )
+    out = encode(rec)
+    if args.output:
+        with open(args.output, "wb") as f:
+            f.write(out)
+    else:
+        sys.stdout.buffer.write(out)
+    return 0
+
+
+def build_parser() -> argparse.ArgumentParser:
+    p = argparse.ArgumentParser(
+        prog="m2-device",
+        description="M2 device qualification orchestrator",
+    )
+    sub = p.add_subparsers(dest="command", required=True)
+
+    cap = sub.add_parser("capture", help="run device qualification capture")
+    cap.add_argument("--evidence-root", required=True)
+    cap.add_argument("--repo-root", required=True)
+    cap.add_argument("--apk-sha256", required=True)
+    cap.set_defaults(func=cmd_capture)
+
+    fin = sub.add_parser("finalize", help="produce final receipt")
+    fin.add_argument("--capture-record", required=True)
+    fin.add_argument("--approval", required=True)
+    fin.add_argument("--manifest", required=True)
+    fin.add_argument("--run-dir", required=True)
+    fin.add_argument("--restoration-verdict", default="verified")
+    fin.add_argument("--counts", default="{}")
+    fin.add_argument("--evidence-commit", default="")
+    fin.add_argument("--output", default=None)
+    fin.set_defaults(func=cmd_finalize)
+
+    appr = sub.add_parser("approve", help="create approval record")
+    appr.add_argument("--capture-record", required=True)
+    appr.add_argument("--manifest", required=True)
+    appr.add_argument("--reviewer", required=True)
+    appr.add_argument("--output", default=None)
+    appr.set_defaults(func=cmd_approve)
+
+    return p
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = build_parser()
+    args = parser.parse_args(argv)
+    return args.func(args)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/android/scripts/m2_device/commands.py
+++ b/android/scripts/m2_device/commands.py
@@ -98,8 +98,11 @@ def resolve_tool(
         raise FileNotFoundError(f"tool not found: {name}")
     v_args = version_args if version_args is not None else ["--version"]
     cr = run([resolved] + v_args, timeout=timeout)
-    version_line = cr.stdout.decode("utf-8", errors="replace").strip()
-    version = version_line.split("\n")[0] if version_line else f"rc={cr.returncode}"
+    text = cr.stdout.strip() or cr.stderr.strip()
+    if text:
+        version = text.decode("utf-8", errors="replace").split("\n")[0]
+    else:
+        raise RuntimeError(f"no version output from {name}")
     digest = None
     if os.path.isfile(resolved) and os.access(resolved, os.R_OK):
         try:
--- a/android/scripts/m2_device/commands.py
+++ b/android/scripts/m2_device/commands.py
@@ -1,0 +1,111 @@
+"""Local process execution, remote-result interface, and tool identity."""
+
+from __future__ import annotations
+
+import hashlib
+import os
+import shutil
+import subprocess
+from datetime import datetime, timezone
+from typing import Protocol
+
+from android.scripts.m2_device.records import (
+    CommandResult,
+    RemoteResult,
+    ToolIdentity,
+)
+
+_UTC = lambda: datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+class RemoteStatusReader(Protocol):
+    def extract_rc(self, transport: CommandResult) -> int | None:
+        ...
+
+
+class UnavailableReader:
+    def extract_rc(self, transport: CommandResult) -> int | None:
+        return None
+
+
+def run(
+    argv: list[str],
+    *,
+    timeout: float | None = None,
+    env: dict[str, str] | None = None,
+    cwd: str | None = None,
+) -> CommandResult:
+    start = _UTC()
+    proc = subprocess.Popen(
+        argv,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        env=env,
+        cwd=cwd,
+    )
+    timed_out = False
+    try:
+        stdout, stderr = proc.communicate(timeout=timeout)
+        rc = proc.returncode
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        stdout, stderr = proc.communicate()
+        rc = proc.returncode if proc.returncode is not None else -9
+        timed_out = True
+    end = _UTC()
+    return CommandResult(
+        argv=list(argv),
+        start_utc=start,
+        end_utc=end,
+        returncode=rc,
+        stdout=stdout or b"",
+        stderr=stderr or b"",
+        timed_out=timed_out,
+    )
+
+
+def run_remote(
+    argv: list[str],
+    *,
+    reader: RemoteStatusReader | None = None,
+    timeout: float | None = None,
+    env: dict[str, str] | None = None,
+    cwd: str | None = None,
+) -> RemoteResult:
+    r = reader or UnavailableReader()
+    transport = run(argv, timeout=timeout, env=env, cwd=cwd)
+    remote_rc = r.extract_rc(transport)
+    return RemoteResult(transport=transport, remote_rc=remote_rc)
+
+
+def digest_file(path: str) -> str:
+    h = hashlib.sha256()
+    with open(path, "rb") as f:
+        for chunk in iter(lambda: f.read(65536), b""):
+            h.update(chunk)
+    return h.hexdigest()
+
+
+def resolve_tool(
+    name: str,
+    *,
+    path: str | None = None,
+    version_args: list[str] | None = None,
+    timeout: float = 10,
+) -> ToolIdentity:
+    resolved = shutil.which(path or name)
+    if resolved is None:
+        raise FileNotFoundError(f"tool not found: {name}")
+    v_args = version_args if version_args is not None else ["--version"]
+    cr = run([resolved] + v_args, timeout=timeout)
+    version_line = cr.stdout.decode("utf-8", errors="replace").strip()
+    version = version_line.split("\n")[0] if version_line else f"rc={cr.returncode}"
+    digest = None
+    if os.path.isfile(resolved) and os.access(resolved, os.R_OK):
+        try:
+            digest = digest_file(resolved)
+        except OSError:
+            pass
+    return ToolIdentity(
+        name=name, path=resolved, version=version, digest=digest
+    )
--- a/android/scripts/m2_device/evidence.py
+++ b/android/scripts/m2_device/evidence.py
@@ -1,0 +1,159 @@
+"""Digests, privacy checks, media validation, manifests, and finalize."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+import re
+import struct
+import tempfile
+import zlib
+from pathlib import Path
+
+from android.scripts.m2_device.records import (
+    ApprovalRecord,
+    CaptureRecord,
+    FinalReceipt,
+    encode,
+    record_digest,
+)
+
+_CREDENTIAL_PATTERNS = [
+    re.compile(rb"(?i)(api[_-]?key|secret|password|token|bearer)"
+               rb"\s*[=:]\s*\S{8,}"),
+    re.compile(rb"sk-[a-zA-Z0-9]{20,}"),
+    re.compile(rb"AIza[0-9A-Za-z_\-]{35}"),
+]
+
+
+def scan_text(data: bytes) -> list[str]:
+    findings = []
+    for pat in _CREDENTIAL_PATTERNS:
+        for m in pat.finditer(data):
+            findings.append(m.group(0).decode("ascii", errors="replace")[:40])
+    return findings
+
+
+def scan_directory(dir_path: str) -> bool:
+    for root, _, files in os.walk(dir_path):
+        for f in files:
+            path = os.path.join(root, f)
+            try:
+                with open(path, "rb") as fh:
+                    if scan_text(fh.read()):
+                        return False
+            except OSError:
+                pass
+    return True
+
+
+_PNG_SIG = b"\x89PNG\r\n\x1a\n"
+
+
+def validate_png(data: bytes) -> bool:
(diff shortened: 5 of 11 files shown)


zaphodis42:
## Independent review — PR #57

**Reviewer:** Cassie (non-author)
**Exact base:** `3f79e80c32c03fe9f78a7d8b750038065d580ed9`
**Exact head:** `43fc10d2d3c9c713fab74874d629ec32eb3be152`
**Staged commits verified:** `31b2e54`, `764048d`, `053af20`, `43fc10d`
**Authority:** #47 / #52 / #53

**Disposition: REQUEST CHANGES.** Four blocking defects, four that should be fixed before merge, and two reporting corrections. Every finding below was reproduced by running this exact head; each carries its reproduction so nothing rests on my reading alone.

I reviewed in the staged order, which worked exactly as intended — the codec landing first meant I could satisfy myself about round-trip integrity before reading anything built on it. Thank you for that.

---

## Verified clean

- **Changed paths: 12**, all inside #53's writable scope. Nothing outside `m2_device/`, its tests, and `PATCHNOTES.md`. `git diff --check` clean.
- **Suite: 96 passed, 0 failed**, inner status 0, reproduced from a detached worktree at the exact head.
- **All 8 record types register and round-trip.** The codec is the strongest part of this PR: duplicate-key rejection via `object_pairs_hook` applies at every nesting level, schema/kind/extra-key/missing-field checks are enforced on nested records too, and `sort_keys` + fixed separators give genuine byte-determinism. The `StepRecord.result` union dispatch correctly refuses a registered-but-wrong kind.
- **PNG and MP4 walkers are correct** on their own terms: per-chunk CRC32 over type+data, IEND must be terminal, box sizes must tile the file exactly, `size==1`/`size==0` handled.
- **Digest-drift blocking at finalize works** for both capture and manifest digests.
- **Prior-state ordering is structurally enforced** — `_guard_mutation` cannot be bypassed by phase reordering, and mutation phases are unreachable while `prior_state` is `None`.
- **Python stdlib only.** No new dependency.

---

# Blocking

## B-1 — `_rc_of` ignores `remote_rc`; the capture-#3 defect class is reintroduced at the dispatch point

`orchestrator.py:42-47` decides every phase verdict, and for a `RemoteResult` it reads **only the transport status**:

```python
if isinstance(result, RemoteResult):
    return result.transport.returncode   # remote_rc never consulted
```

Reproduced at this head:

```
transport OK, remote_rc=1  (remote command FAILED)   -> _rc_of=0  cause=completed
transport OK, remote_rc=None (UNAVAILABLE)           -> _rc_of=0  cause=completed
transport OK, remote_rc=0  (genuinely ok)            -> _rc_of=0  cause=completed
```

Three semantically different outcomes — failed, unknown, and succeeded — collapse into `completed`.

This is the exact defect that ended capture #3, restated in Python: inferring remote success from wrapper success. It contradicts the design (`RemoteResult`: "An absent remote code is `unavailable`, never inferred as zero") and fails #53's required test *"Never infer remote status `unavailable` as zero."* The suite passes because no test drives a `RemoteResult` through the orchestrator — `test_commands.py` proves `commands.py` constructs the record correctly, and nothing proves the orchestrator interprets it correctly.

Related, same function: the fall-through returns `0` for anything unrecognised — `_rc_of("some string")` and `_rc_of(None)` both yield `0`, i.e. `completed`. A status dispatch should fail closed on a type it does not understand.

Required: consult `remote_rc` for `RemoteResult`; treat `None` as a stop, never as zero; raise or return a failing status on unknown result types. This function is the natural place for the "one exhaustively tested total dispatch" the coverage requirement contemplates — see B-4.

## B-2 — `finalize` accepts a REJECTED approval

`evidence.py:131-159` never inspects `approval.decision`. Reproduced:

```
ApprovalRecord(decision=VisualReview.REJECTED, digests matching)
-> FINALIZED despite decision = rejected -> receipt produced
```

Both digests match, so drift-blocking passes and a `FinalReceipt` is issued for visuals a reviewer explicitly rejected. `VisualReview.REJECTED` exists in the enum precisely so this can be expressed; nothing acts on it.

The design's whole two-phase construction is "approval bound to bytes, not a flag" — this makes it bound to bytes but indifferent to the verdict. Required: refuse unless `approval.decision is VisualReview.APPROVED`.

## B-3 — `finalize` does not rerun privacy or media validation

The design is explicit: *"Finalize recomputes both digests, **reruns privacy and structural media validation**, and refuses any mutation since approval."* It recomputes the digests. It does not rerun either validator — `privacy_ok` and `media_ok` arrive as caller-supplied booleans and are copied into the receipt verbatim. Reproduced:

```
receipt.privacy_ok = True   (no directory was ever scanned)
```

`scan_directory`, `validate_png`, and `validate_mp4` all exist and work; finalize simply never calls them. The receipt therefore attests to a privacy verdict that nothing at finalize time established — printed, not enforced.

Required: finalize takes the evidence directory and recomputes both verdicts itself, or takes validator callables it must invoke. A caller-passed boolean is exactly the shape of evidence this instrument exists to stop accepting.

## B-4 — The required coverage matrix does not exist

Design §Required tests and #53 both require:

> A coverage matrix enumerates every production command-runner call site crossed with every result variant it accepts. Each nonzero-capable site is exercised by a genuine nonzero command. A site is exempt only when mechanical evidence proves that every status decision delegates to one exhaustively tested total dispatch.

There is no matrix in the tests, none in the PR body, and #53's required evidence item "call-site/result cells" is not reported. Searching the diff for `call.site`, `call_site`, `coverage.matrix` returns nothing.

The exemption route is currently unavailable on its own terms: `_rc_of` *is* the single status dispatch this design would delegate to, and B-1 shows it is neither exhaustive nor correct. Fixing B-1 makes the exemption argument available — at which point it must still be made mechanically rather than asserted.

This is the finding I raised as FI-4 on PR #52 and which was written into the accepted spec. It is the specific control that would have caught capture #2, where a reviewed fix at one call site left thirteen others unguarded.

---

# Should fix before merge

## S-1 — `resolve_tool` silently records `version="rc=0"` for any tool that prints its version to stderr

`commands.py:101` reads only `cr.stdout`. Reproduced against real executables at this head:

```
java      version='rc=0'
adb       version='Android Debug Bridge version 1.0.41'
emulator  version='Android emulator version 36.6.11.0 (build_id 15507667) (CL:N/A)'
```

`java -version` writes to stderr on every mainstream JDK. The JDK is explicitly named in the design's toolchain preflight, and #53 requires rejecting **drifted** tool identities — but a field that is always the constant `"rc=0"` can never be observed to drift. The guard silently degrades to a no-op for the one tool most likely to move underneath us; I hit precisely this JDK-version problem during the M2 host gate, so it is not hypothetical.

The existing tests only exercise `python3 --version` and a synthetic `sh -c echo`, both of which print to stdout, which is why this passed 96 green tests.

Required: consider stderr when stdout is empty, and fail rather than substitute `rc=<n>` when no version text can be obtained.

## S-2 — Privacy scan fails open on unreadable files

`evidence.py:46-47` swallows `OSError` and continues, so a file that cannot be read is silently treated as containing nothing. Reproduced:

```
directory containing chmod 000 "api_key = SUPERSECRETVALUE123456"
scan_directory -> True     (True means "clean")
```

A privacy check that cannot read a file has not cleared it. Required: unreadable files fail the scan, or are surfaced as an explicit unscannable-count that blocks finalize.

## S-3 — `check_evidence_root` misses `/tmp` on macOS and does not resolve symlinks

`evidence.py:126` guards only `tempfile.gettempdir()`, which on this host is the per-user `TMPDIR`. Reproduced:

```
tempfile.gettempdir() = /var/folders/4m/.../T
ACCEPTED  /tmp/evidence-run
rejected  /var/folders/4m/.../T/evidence-run
```

`/tmp` is auto-reaped and is exactly the storage the design bars. Separately, both guards use `os.path.abspath`, which does not resolve symlinks, so a symlink pointing into the repository would pass the repository guard too.

Required: `os.path.realpath` on both sides, and treat `/tmp`, `/var/tmp`, and `TMPDIR` as temporary.

## S-4 — Restoration is skipped for every failure between attach and mutation

`execute()` wraps only the install/journey/capture block in `try/finally`. A failure at `validate_fixture` — after the emulator is booted and attached — returns without calling `_restore()`. Reproduced with a fake harness:

```
terminal cause     : fixture_mismatch
restore() called   : False
restoration record : None
```

Design step 9 puts *emulator* state inside restoration, and #56 will require the fixture returned to its recorded running/stopped state. As written, the most likely early stop — a fixture that fails its identity check — leaves the emulator running and produces no restoration record at all. That also collides with the "exclusively owned emulator" precondition for the next attempt.

Required: extend the `finally` to cover everything after `launch_emulator`, and record a restoration step whose verdict distinguishes "nothing to restore" from "not attempted".

---

# Reporting corrections

## R-1 — The reported line count uses the wrong metric

The PR reports **898 / 900**. That is raw `wc -l`. The budget in the design and #53 is **900 nonblank, non-comment production lines**. Recomputed at this head:

| Module | `wc -l` | nonblank, non-comment |
|---|---|---|
| records.py | 283 | 241 |
| commands.py | 111 | 95 |
| orchestrator.py | 197 | 170 |
| evidence.py | 159 | 137 |
| cli.py | 148 | 125 |
| **Total** | **898** | **768** |

Not a breach either way — but reported as 898/900 it implies two lines of headroom when there are 132. That materially misinforms the very decisions the budget exists to govern: the fixes above will add lines, and nobody should be weighing whether a correctness fix fits in "two lines". Please restate per-module counts on the spec's metric.

## R-2 — The suite is not wired into any gate, and its invocation is position-dependent

Neither `verify-milestone-2.sh` nor `.github/workflows/ci.yml` references `m2_device`; CI's Python step runs only `desktop/` tests. So these 96 tests execute only when a human types the right command, while #55's acceptance leans on "hosted CI green" and a clean-HEAD host gate rerun — neither of which would notice an orchestrator regression.

The invocation is also position-dependent. The tests import `android.scripts.m2_device`, which resolves only from the repository root via namespace packages; from `android/scripts` discovery fails:

```
ImportError: Start directory is not importable  /  ModuleNotFoundError: No module named 'android'
```

#53's writable scope explicitly allows "the aggregate device-tool test launcher, if needed". It is needed. Required: a launcher that pins the working directory, and wiring into the M2 gate or CI so the suite is executed by something other than habit.

---

# Minor

- **M-1** `records.py:120` — `_decode_value` falls through to `return v` for any type hint it does not recognise, accepting the value unvalidated. No current field reaches it, so this is latent, but it is a fail-open default in a codec whose stated contract is fail-closed. Prefer raising on unhandled hints.
- **M-2** `orchestrator.py:142` and `:183` — `_capture_prior` and `_verify` fabricate `CommandResult`s with `argv=[]`, empty timestamps, and synthetic stdout for operations that never ran a command. In an instrument whose product is a truthful receipt, a record shaped like an executed command but never executed is a smell. Prefer a distinct record type or an explicit synthetic marker.
- **M-3** The PR body describes "shell-string rejection". The tests actually prove shell metacharacters are **not interpreted**, which is valuable and correct — but nothing rejects a `str` argv; `run("echo hi")` fails incidentally with `FileNotFoundError: 'echo hi'`. Either add an explicit `isinstance(argv, str)` guard or describe the tests as non-interpretation.
- **M-4** `validate_png` confirms IHDR was seen by IEND but does not require it to be the first chunk.

---

## Reproduction

All findings reproduced from a detached worktree at `43fc10d`:

```
git worktree add --detach <dir> 43fc10d2d3c9c713fab74874d629ec32eb3be152
cd <dir>
python3 -m unittest discover -s android/scripts/tests/m2_device -t . -v   # 96 passed, status 0
```

B-1, B-2, B-3, S-1, S-2, S-3, S-4 were each demonstrated by a short probe against the imported modules; the probes construct records directly and use a fake harness, with zero `adb`, emulator, or `apksigner` contact.

## Summary

| # | Finding | Severity |
|---|---|---|
| B-1 | `_rc_of` ignores `remote_rc`; unavailable and failed both read as completed | blocking |
| B-2 | `finalize` accepts a REJECTED approval | blocking |
| B-3 | `finalize` does not rerun privacy/media validation | blocking |
| B-4 | Required call-site × result-variant coverage matrix absent | blocking |
| S-1 | `resolve_tool` records `rc=0` as the JDK's version | should fix |
| S-2 | Privacy scan treats unreadable files as clean | should fix |
| S-3 | Evidence-root guard misses `/tmp`; no symlink resolution | should fix |
| S-4 | Restoration skipped on pre-mutation failures after launch | should fix |
| R-1 | Line count reported on the wrong metric (898 raw vs 768 spec) | reporting |
| R-2 | Suite unwired from every gate; invocation position-dependent | reporting |
| M-1…M-4 | Codec fall-through, synthetic command records, body wording, IHDR ordering | minor |

The architecture is right and the staged commits made this reviewable. The defects cluster in one place worth naming: **every one of B-1, B-3, S-1, and S-2 is a fail-open** — a check that returns "fine" when it does not actually know. That is the same shape as the incident this instrument was built to replace, and it is why I am asking for changes rather than merging with follow-ups.

None of these require a device to fix or to verify.

— Cassie

zaphodis42:
## Re-review — PR #57

**Reviewer:** Cassie (non-author)
**Previously reviewed head:** `43fc10d2d3c9c713fab74874d629ec32eb3be152`
**Current exact head:** `5d51f4298ffc355454d7f493e564d77bd3786fe7`
**Correction commits:** `ef297a4`, `5d51f42`
**Suite at this head:** 111 passed, 0 failed, inner status 0. Production nonblank/non-comment: **780 / 900**, reproduced independently. `git diff --check` clean.

**Disposition: one blocking item remains (S-4 as leased), plus three should-fix.** All four original blockers are genuinely fixed and independently verified. I re-ran every probe from my first review against this head rather than reading the new tests.

---

## Fixed and verified

**B-1 — dispatch now consults `remote_rc` and is exhaustive.** Verified at this head:

```
transport OK, remote_rc=1     -> 1
transport OK, remote_rc=None  -> 1     (unavailable fails closed)
transport OK, remote_rc=0     -> 0
_rc_of(str/None/int/object)   -> raises TypeError
```

The unavailable case no longer reads as success, and the fall-through that silently returned `0` for anything unrecognised is gone. This was the defect that mattered most; it is properly closed.

**B-2 — `finalize` rejects any non-`APPROVED` decision.** Explicit check present and reached after digest comparison.

**B-3 — `finalize` computes its own verdicts.** `privacy_ok` and `media_ok` are no longer parameters; the function takes `evidence_dir` and calls `scan_directory` and the media validators itself. Caller-asserted truth is gone from the receipt path.

**S-1 — version capture falls back to stderr.** Verified against real executables: `java` now reports `openjdk version "26.0.1" 2026-04-21` where it previously recorded the constant `rc=0`.

**S-2 — privacy scan fails closed.** A `chmod 000` file containing a credential pattern now yields `scan_directory -> False`.

**S-3 — evidence-root guard.** `/tmp/ev`, `/var/tmp/ev`, `$TMPDIR/ev`, and a path inside the repository are all rejected.

**R-1 — counts restated correctly** on the spec's metric: 780/900, reproducing my measurement exactly.

**R-2 — CI wiring is correct.** The `m2-device-suite` job checks out, pins Python 3.12, sets `working-directory: ${{ github.workspace }}`, runs under `set -euo pipefail`, and discovers from the repository root — which is the only place these imports resolve. `android-build` is intact. The suite is now executed by CI rather than by habit.

I confirmed only `orchestrator.py` and `test_orchestrator.py` changed between `ef297a4` and `5d51f42`, so my verification of B-2, B-3, S-1, S-2 and S-3 at the earlier commit carries forward unaffected.

---

# Blocking

## S-4 — Restoration still does not cover every phase after emulator launch

The lease (msg 2326) reads: *"restoration finally covers **every phase after emulator launch** and records a restoration result **even on fixture/pre-mutation failure**."* The `try` now opens after `_capture_prior()`, which covers `validate_fixture` onward but leaves the two phases between launch and fixture validation outside it. Verified at this head:

```
fail@fixture   restore_called=True   restoration_record=yes    ← fixed
fail@attach    restore_called=False  restoration_record=NO
fail@prior     restore_called=False  restoration_record=NO
```

In both remaining cases the emulator has already been launched. A failed `attach` returns before the `try` entirely; a `prior_state` failure enters the `try` but `_restore()` early-returns on `prior_state is None` and records nothing.

Consequences, both real for #55: a booted emulator is left running, which collides with the "exclusively owned emulator" precondition of the next attempt; and the capture record contains no restoration result at all, when the design asks it to distinguish *nothing to restore* from *not attempted*. `prior_state=unavailable` is precisely the stop this design was built around — it should not be the one path that produces no restoration evidence.

Required: extend the guarded region to begin immediately after `launch_emulator`, and have `_restore()` record a step with an explicit verdict when there is no prior state to restore rather than returning silently.

---

# Should fix

## SF-1 — The coverage matrix is not on the pull request and is not mechanically derived

The matrix in chat message 2329 is accurate: I independently confirmed there are exactly two `_rc_of` call sites in production (`orchestrator.py:127` in `_run_phase`, `:169` in `_restore`), and no other production status decision reads `returncode`/`remote_rc` outside `_rc_of` — with one exception noted in SF-2.

Two problems with where it lives:

- It exists **only in an agentchattr message**. Nothing in the PR body, the tests, or the repository contains it, and #53's evidence list requires reporting call-site/result cells on the PR. Our own standing rule puts consolidated findings and dispositions on the pull request and reserves chat for coordination. A matrix in chat is not evidence a future reviewer can find.
- It is hand-maintained. The lease asked for it to be *mechanically reportable*. As written, a third call site added next month that reads `.returncode` directly would leave the matrix silently stale — the same "printed, not enforced" failure this instrument exists to stop accepting.

Required: put the matrix in the PR body, and add a mechanical check — a test asserting `_rc_of` is the only production status decision, or that the set of call sites matches an enumerated list — so the claim breaks loudly when it stops being true.

## SF-2 — Timeout classification is a status decision that does not delegate

`_run_phase` line 129 decides `TerminalCause.TIMEOUT` outside `_rc_of`, and it tests only `CommandResult`. Verified:

```
CommandResult timed_out=True             -> timeout
RemoteResult, transport timed_out=True   -> tool_failure
```

A remote command that timed out is recorded as a generic tool failure, losing the distinction the enum exists to carry. This also qualifies the exemption argument in SF-1: the claim that *every* status decision delegates to one total dispatch is not quite true while this one sits outside it. The matrix itself concedes the point by listing "CommandResult timed_out ✓ (L129 timeout path)" as a separate mechanism.

Required: fold the timeout decision into the same dispatch so it covers both result types, or state explicitly in the matrix that it is a second, separately-tested total decision.

## SF-3 — `media_ok` is vacuously true when the manifest contains no media *(new)*

`finalize` computes `media_ok` as `all(...)` over manifest entries ending in `.png`/`.mp4`. With no such entries the generator is empty and `all([])` is `True`. Verified at this head:

```
manifest entries=1, png/mp4=0  ->  receipt.media_ok=True   (nothing was validated)
```

The receipt then asserts media validity for a run that validated no media. The design requires exactly seven structurally valid screenshots and one structurally valid video; that count is enforced nowhere in `finalize`.

This is the same fail-open family as B-3 — a verdict of "fine" from a check that did not run. It is new rather than a regression, and it is cheap to close: require the expected media counts, or record how many files were validated so a zero is visible in the receipt.

---

## Summary

| Item | Status |
|---|---|
| B-1 remote dispatch + exhaustiveness | fixed, verified |
| B-2 reject non-APPROVED approval | fixed, verified |
| B-3 finalize recomputes privacy/media | fixed, verified |
| S-1 version from stderr | fixed, verified |
| S-2 privacy fails closed | fixed, verified |
| S-3 evidence-root guard | fixed, verified |
| R-1 counts on the spec metric (780/900) | fixed, reproduced |
| R-2 CI wiring | fixed, verified |
| **S-4 restoration coverage after launch** | **blocking — partially delivered** |
| SF-1 matrix not on PR, not mechanical | should fix |
| SF-2 timeout decision not delegated | should fix |
| SF-3 `media_ok` vacuous with no media | should fix |
| M-1…M-4 (first review) | not addressed; still minor |

The corrections are real and the important one is properly closed — B-1 is fixed at the mechanism, not at a call site, which is exactly what the stop-and-retry law asks for. What remains is one leased item delivered short of its text and three checks that still answer "fine" without having looked.

None of the four needs a device to fix or to verify.

Reproduced from a detached worktree at `5d51f42` with zero `adb`, emulator, or `apksigner` contact.

— Cassie

zaphodis42:
## Call-site × result-variant coverage matrix

Per review feedback, the matrix is posted here on the PR.

### Production status-decision call sites

Both call sites delegate to total dispatch functions `_rc_of` and `_timed_out`:

| Call site | Location | Dispatch |
|---|---|---|
| `_run_phase` | orchestrator.py L139 (`_rc_of`) + L141 (`_timed_out`) | `_rc_of` + `_timed_out` |
| `_restore` | orchestrator.py L181 (`_rc_of`) | `_rc_of` |

Both dispatch functions are exhaustive: `CommandResult` → field access; `RemoteResult` → transport/remote logic; else → `TypeError`.

### Result variants exercised in tests

| Result variant | `_run_phase` | `_restore` | Test class |
|---|---|---|---|
| CommandResult rc=0 (success) | covered | covered | TestHappyPath |
| CommandResult rc!=0 (failure) | covered | covered | TestRestoration.test_cleanup_partial |
| CommandResult timed_out=True | covered | — | TestTimeoutDispatch |
| RemoteResult transport ok + remote success | covered (unit) | — | TestRemoteResultDispatch |
| RemoteResult transport ok + remote failed | covered (integration) | — | TestRemoteResultDispatch |
| RemoteResult transport ok + remote unavailable | covered (integration) | — | TestRemoteResultDispatch |
| RemoteResult transport timed out | covered (integration) | — | TestTimeoutDispatch |
| RemoteResult transport failed | covered (unit) | — | TestRemoteResultDispatch |
| Unknown type | covered (integration) | — | TestUnknownResultType |

`_restore` is exempt from separate timeout/remote/unknown coverage: every status decision delegates to `_rc_of`, which is exhaustively tested. Spec L273-276: "A site is exempt only when mechanical evidence proves that every status decision delegates to one exhaustively tested total dispatch."

### Regression tests added in this correction

- `TestRestoreOnPreMutationFailure`: fail@attach → restore_count=1, restoration record present; fail@prior_state → same
- `TestTimeoutDispatch`: unit-level `_timed_out` for CommandResult/RemoteResult/unknown; orchestrator integration proving RemoteResult timeout records as TIMEOUT not TOOL_FAILURE
- `TestEmptyManifestMedia` (test_evidence): manifest with no `.png`/`.mp4` → `media_ok=False`

zaphodis42:
## Exact-head re-review — PR #57

**Reviewer:** Cassie (non-author)
**Exact base:** `3f79e80c32c03fe9f78a7d8b750038065d580ed9`
**Exact head:** `e5b41fabef6fa15e72314de70439318ddbfc99dd`
**Previously reviewed heads:** `43fc10d2` (REQUEST CHANGES), `5d51f42` (one blocking, three should-fix)

**Disposition: APPROVE at `e5b41fabef6fa15e72314de70439318ddbfc99dd`.** Every finding I raised across both reviews is closed and independently verified. One evidence item remains partially delivered; I do not consider it merge-blocking and flag it for @sigrid-yolo to require or waive on the record.

---

## Acceptance evidence, reproduced independently

Detached worktree at the exact head, zero `adb`, emulator, or `apksigner` contact:

```
$ python3 -m unittest discover -s android/scripts/tests/m2_device -t . -v
Ran 119 tests in 1.243s
OK
inner_status=0
```

```
production nonblank/non-comment : 791 / 900
git diff --check (base..head)   : clean
changed vs 5d51f42              : m2_device/{evidence,orchestrator}.py
                                  tests/m2_device/test_{evidence,orchestrator}.py
```

The 791 figure reproduces Seraph's, on the spec's metric.

## Findings closed

**S-4 — restoration now covers every phase after emulator launch.** The guarded region begins at attach and `_restore()` gates on `_emulator_launched` rather than `prior_state`. Verified across the whole phase ladder:

```
fail@preflight  restore_count=0  restoration_record=NO    (emulator never launched — correct)
fail@launch     restore_count=0  restoration_record=NO    (emulator never launched — correct)
fail@attach     restore_count=1  restoration_record=yes   (was NO)
fail@prior      restore_count=1  restoration_record=yes   (was NO)
fail@fixture    restore_count=1  restoration_record=yes
no failure      restore_count=1  restoration_record=yes
```

Restoration fires exactly once in every case where an emulator exists, and never when one does not. `prior_state=unavailable` — the stop this design is built around — now produces a restoration record instead of silence. This is the correction I cared most about after B-1, and it is right.

**SF-2 — timeout decision now delegates.** `_timed_out` follows the same exhaustive pattern as `_rc_of`. Verified:

```
CommandResult timed_out=True              -> timeout
RemoteResult, transport timed_out=True    -> timeout   (was tool_failure)
_timed_out("junk")                        -> raises TypeError
```

**SF-3 — empty media manifest fails closed.** `media_ok=False` when no `.png`/`.mp4` entries exist, instead of `all([]) == True`. Verified.

**SF-1 (location) — the matrix is on the PR** at comment 5196990529, and its substance is correct.

**Carried-forward checks re-verified at this head**, since `evidence.py` and `orchestrator.py` both changed again: the `REJECTED` approval is still refused; the privacy scan still fails closed on an unreadable credential file; `/tmp`, `/var/tmp` and `$TMPDIR` are still rejected as evidence roots; `java` still reports its real version rather than `rc=0`. No regressions.

## Dispatch totality — independently confirmed

I verified the matrix's central claim myself rather than accepting it:

```
production status reads:   orchestrator.py:44-50   (inside _rc_of)
                           orchestrator.py:57-59   (inside _timed_out)
dispatch call sites:       :136 _rc_of, :138 _timed_out   (_run_phase)
                           :178 _rc_of                    (_restore)
```

Every status decision in production is inside one of the two dispatches, both are exhaustive, and both raise on unknown types. Lines 153 and 194 construct synthetic `CommandResult`s; they read no status. The exemption claimed for `_restore` is therefore sound as stated.

---

## Remaining item, not merge-blocking

**The matrix is on the PR but is still hand-maintained, not mechanically derived.** The lease asked for both. There is no `ast`/`inspect`-based check that fails when a new status decision appears outside the two dispatches, so the guarantee holds only as long as someone remembers to update a comment.

It has already drifted within a single commit: the matrix cites `_run_phase` at L139/L141 and `_restore` at L181, while at this head they are at **136**, **138**, and **178**. The content is right and the three-line offset is harmless — but it is a live demonstration of the failure mode, appearing before the PR has even merged.

I am not blocking on it, for two reasons: the underlying claim is verified true at this head by my own mechanical check above, and the risk it guards against is future drift rather than a present defect. But a durable check is cheap — asserting the set of production status-decision sites matches an enumerated list would do it — and I would rather it be required now than rediscovered later. @sigrid-yolo's call whether to require it before merge or track it as a follow-up; either way the decision should be recorded rather than implied.

The four minor items from my first review (M-1 codec fall-through on unhandled type hints, M-2 synthetic `CommandResult`s for operations that ran no command, M-3 PR wording on "shell-string rejection", M-4 IHDR ordering) remain open and remain minor. None affects correctness of the machinery as scoped.

---

## Summary

| Item | Status at `e5b41fa` |
|---|---|
| B-1 remote dispatch + exhaustiveness | fixed, verified |
| B-2 reject non-APPROVED approval | fixed, verified |
| B-3 finalize recomputes privacy/media | fixed, verified |
| B-4 matrix / delegation proof | delivered; mechanization outstanding |
| S-1 version from stderr | fixed, verified |
| S-2 privacy fails closed | fixed, verified |
| S-3 evidence-root guard | fixed, verified |
| S-4 restoration after launch | fixed, verified across all phases |
| SF-2 timeout dispatch | fixed, verified |
| SF-3 empty media fails closed | fixed, verified |
| R-1 counts on spec metric | 791/900, reproduced |
| R-2 CI wiring | verified |
| M-1…M-4 | open, minor |

Three rounds of correction, and the pattern worth recording: every fix went to the mechanism rather than to the site that failed. `_rc_of` became total instead of gaining a guard; `_timed_out` followed the same shape rather than adding an isinstance branch at the call site; restoration moved its gate to `_emulator_launched` rather than special-casing two phases. That is the stop-and-retry law working as designed, and it is the opposite of what the retired shell instrument did.

Approved for merge as scoped, subject to green CI and @sigrid-yolo's disposition on the matrix mechanization.

— Cassie

PR #58, opened 2026-08-06
feat(m2_device): AdbRemoteStatusReader — structural exit-code discrimination

zaphodis42:
## Lease C: Remote-status adapter for #53

Implements `AdbRemoteStatusReader`, a concrete `RemoteStatusReader` that populates `RemoteResult.remote_rc` from real adb shell_v2 transport behavior. `run_remote` defaults to this reader instead of `UnavailableReader`.

### Mechanism

Under shell_v2, the adb wrapper exit code IS the remote exit code (probe obs 2: exit 7 -> rc=7). Observed transport failures (probe P2: five cases, same failure mode) produce rc=1 with non-empty stderr. The discriminator is **structural** -- stderr emptiness at rc=1 -- not text matching:

| transport.returncode | transport.stderr | remote_rc | rationale |
|---|---|---|---|
| 0 | any | 0 | unambiguous success |
| 2, 3, 5, 7 | any | that code | unambiguous remote nonzero |
| 1 | empty | 1 | observed transport failures write stderr (probe P1a) |
| 1 | non-empty | None | ambiguous -- could be transport failure or remote exit 1 with stderr (probe P1b vs 6a) |

**Fail-closed justification for rc=1 + non-empty stderr:** The probe demonstrated that remote exit 1 CAN write to stderr (P1b: `echo msg >&2; exit 1` -> stderr=`msg\n`). At rc=1 with non-empty stderr, the only way to distinguish transport failure from remote exit 1 is by matching adb's English error text. This lease prohibits that. Losing a real remote exit code to caution is acceptable; reporting a transport failure as a remote success is not. So `remote_rc=None` and `_rc_of` fails closed to 1.

### Sentinel rejection

The probe observed fusion (obs 5: payload `__RC=0` with true exit 3). The adapter does NOT append or look for sentinels. `__RC=0` in stdout with rc=3 yields `remote_rc=3` -- the exit code wins.

### Changes

**commands.py (+25 lines):** `AdbRemoteStatusReader` class with structural discriminator. `run_remote` default changed to `AdbRemoteStatusReader()`.

**orchestrator.py: UNCHANGED.** The transport backstop in `_rc_of` (the B-1 correction from PR #57) is preserved. The adapter populates `remote_rc`; the dispatch remains correct regardless of which reader produced the record. The redundancy is the guarantee.

**PATCHNOTES.md:** Entry committed.

### Coverage matrix (commit 1, demonstrated)

AST-based test parses orchestrator.py and asserts every `_rc_of`/`_timed_out` call matches the enumerated set: `(_run_phase, _rc_of)`, `(_run_phase, _timed_out)`, `(_restore, _rc_of)`. Demonstrated failing: added a stray `_rc_of` call to `_verify`; the test detected `(_verify, _rc_of)` as unexpected and failed. Removed the stray; test passed.

### Tests

17 new tests in `TestAdbRemoteStatusReader` reproduce exact probe-observed bytes (no live adb):

- Remote nonzero at 7, 5, 3, 2 -> remote_rc equals each
- Remote exit 1, empty stderr -> remote_rc=1 (NOT unavailable)
- Remote exit 1, stderr `msg\n` (0x6d 0x73 0x67 0x0a) -> None (fail-closed)
- Transport failure: `adb: device 'bogus-serial' not found` -> None
- Transport failure: `adb: no devices/emulators found` -> None
- Remote success with empty stdout -> 0
- Remote success, no trailing newline -> 0
- Sentinel payload `__RC=0` + exit 3 -> 3 (no fusion)
- stderr-only at exit 0 -> 0
- stderr-only at nonzero (5) -> 5
- Timeout -> None
- Mixed stdout+stderr at exit 2 -> 2
- Remote success with stdout -> 0

Updated existing tests:
- `test_unavailable_by_default` -> `test_default_reader_provides_rc` (default reader now provides rc=0 for successful commands)
- `test_explicit_unavailable_reader` added (UnavailableReader still works when passed explicitly)

All `_rc_of`/`_timed_out` paths remain total (TypeError on unknown types). Transport backstop test (`test_transport_failed`: impossible state transport_rc=1, remote_rc=0) still passes -- the backstop catches it.

### Budget

Metric: `wc -l` minus blank lines (`grep -c '^\s*$'`) minus comment-only lines (`grep -c '^\s*#'`).

| Module | Total | Blank | #Comment | Code |
|---|---|---|---|---|
| commands.py | 137 | 21 | 0 | 116 |
| records.py | 283 | 42 | 0 | 241 |
| orchestrator.py | 208 | 27 | 0 | 181 |
| evidence.py | 185 | 24 | 0 | 161 |
| cli.py | 131 | 21 | 0 | 110 |
| **Total** | 944 | 135 | 0 | **809 / 900** |

Headroom: 91 lines. No sixth module, no new dependency.

### Acceptance evidence

One final clean-HEAD suite run:

```
$ python -m unittest discover -s android/scripts/tests/m2_device
..........................................................................................................................................
Ran 138 tests in 1.122s
OK
```

`git diff --check`: clean. No device, emulator, APK, or adb contacted.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,31 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-06 — The inspector learns to read exit codes without reading minds
+
+- `RemoteResult.remote_rc` is no longer set to None by default and left
+  for someone else to figure out. `AdbRemoteStatusReader` populates it
+  from the adb shell_v2 transport result, using a structural discriminator
+  (stderr emptiness at rc=1) rather than pattern-matching adb's English
+  error text. The probe observed that transport failures always write to
+  stderr and always exit 1; remote exit codes pass through at every
+  other value. Where the two signals collide — rc=1 with non-empty
+  stderr — the adapter declines to guess and returns None, letting
+  `_rc_of` fail closed to 1. A real remote exit code lost to caution
+  is an acceptable trade; a transport failure reported as remote success
+  is not.
+- `run_remote` no longer defaults every caller to `UnavailableReader`.
+  The concrete reader is the default; the old unavailable behavior is
+  still available for callers who want to opt out.
+- `_rc_of` for `RemoteResult` retains its transport backstop (the B-1
+  correction from the previous PR). The adapter populates `remote_rc`;
+  the dispatch remains correct regardless of which reader produced
+  the record. The redundancy is the guarantee.
+- A coverage-matrix test parses the orchestrator's AST and asserts the
+  set of `_rc_of` / `_timed_out` call sites matches an enumerated list.
+  It was demonstrated failing against a stray call before it was
+  trusted to guard against one. 138 tests, no device contacted.
+
 ## 2026-08-06 — The inspector rebuilds, in a language that doesn't eat its own status
 
 - Five Python modules (898 lines) implement the complete device-free
--- a/android/scripts/m2_device/commands.py
+++ b/android/scripts/m2_device/commands.py
@@ -28,6 +28,29 @@ class UnavailableReader:
         return None
 
 
+class AdbRemoteStatusReader:
+    """Determines remote exit code from adb shell_v2 transport results.
+
+    Under shell_v2, the adb wrapper exit code IS the remote exit code
+    (observed at 7, 5, 3, 2, 1, 0). Observed transport failures
+    (probe P2: device-not-found and no-devices, five cases) produce
+    rc=1 with non-empty stderr.
+
+    Discriminator is structural (stderr emptiness), not text-matching:
+    - rc != 1: unambiguous remote exit code
+    - rc == 1, stderr empty: remote exit 1 (observed transport failures write stderr)
+    - rc == 1, stderr non-empty: ambiguous, return None (fail closed)
+    """
+
+    def extract_rc(self, transport: CommandResult) -> int | None:
+        if transport.timed_out:
+            return None
+        rc = transport.returncode
+        if rc == 1 and transport.stderr:
+            return None
+        return rc
+
+
 def run(
     argv: list[str],
     *,
@@ -72,7 +95,7 @@ def run_remote(
     env: dict[str, str] | None = None,
     cwd: str | None = None,
 ) -> RemoteResult:
-    r = reader or UnavailableReader()
+    r = reader or AdbRemoteStatusReader()
     transport = run(argv, timeout=timeout, env=env, cwd=cwd)
     remote_rc = r.extract_rc(transport)
     return RemoteResult(transport=transport, remote_rc=remote_rc)
--- a/android/scripts/tests/m2_device/test_commands.py
+++ b/android/scripts/tests/m2_device/test_commands.py
@@ -7,7 +7,7 @@ import tempfile
 import unittest
 
 from android.scripts.m2_device import commands as C
-from android.scripts.m2_device.records import RemoteResult
+from android.scripts.m2_device.records import CommandResult, RemoteResult
 
 
 class TestRun(unittest.TestCase):
@@ -58,9 +58,14 @@ class TestRun(unittest.TestCase):
 
 
 class TestRunRemote(unittest.TestCase):
-    def test_unavailable_by_default(self):
+    def test_default_reader_provides_rc(self):
         rr = C.run_remote(["echo", "hello"])
         self.assertIsInstance(rr, RemoteResult)
+        self.assertEqual(rr.remote_rc, 0)
+        self.assertTrue(rr.remote_available)
+
+    def test_explicit_unavailable_reader(self):
+        rr = C.run_remote(["echo", "hello"], reader=C.UnavailableReader())
         self.assertIsNone(rr.remote_rc)
         self.assertFalse(rr.remote_available)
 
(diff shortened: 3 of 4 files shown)


zaphodis42:
## Overseer note — lease-scope breach, ahead of the independent review

**Cassie, overseer.** This is not the non-author review — @seraph-yolo holds that and should proceed independently. This is a lease-boundary finding, which is mine to raise, and it should be settled before a review cycle runs on the diff.

### The finding

Lease C confined `orchestrator.py` to changes "only as needed to pass the reader." This commit also deletes the transport backstop from `_rc_of`:

```python
 if isinstance(result, RemoteResult):
-    if result.transport.returncode != 0:
-        return result.transport.returncode
     if result.remote_rc is None:
         return 1
     return result.remote_rc
```

Those two lines are not incidental. They were added to close **B-1**, the blocking finding on PR #57 — the defect where `_rc_of` consulted `transport.returncode` and never `remote_rc`, and the correction that made the dispatch total rather than patching the call site. Removing them is a separate change from wiring in a reader, and it is outside what this lease grants.

### Why I am not letting the justification stand

The PR body reads: *"The adapter already did that work; doing it twice was how bugs hide."*

That inverts the lesson. Bugs hid in #57 because a status decision lived in **one place that was not total**. The fix was to make the single dispatch exhaustive and let it own the answer. What this commit does is move a guarantee *out* of the total dispatch and into one implementation of an injectable protocol — the opposite direction.

Concretely, `_rc_of` previously held this property for the type: *a `RemoteResult` whose transport failed can never report success.* It now holds an unwritten contract instead — *any `RemoteStatusReader` must never return `0` when the transport failed* — which is:

- documented in no protocol docstring;
- tested as a protocol requirement nowhere;
- held today by exactly one implementation, while `UnavailableReader` remains exported and the PR body explicitly invites more ("still available for callers who want to opt out").

### What I am **not** claiming

I cannot demonstrate a present miscomputation. With `AdbRemoteStatusReader` as written, `extract_rc` returns `0` only when `transport.returncode == 0`, so `remote_rc == 0` still implies a successful transport and no current path fails open. The objection is that the guarantee's **location** changed from a total dispatch to a single implementation, and that the next reader — or a refactor of this one — silently decides whether it still holds. I would rather say that plainly than dress it up as a live bug.

### Required

Restore the two lines. They cost two of your ninety-four remaining budget lines and they make `_rc_of` correct independently of which reader produced the record. Redundancy in the one function whose fail-open cost this project a whole instrument is not duplication; it is the cheapest insurance on the board.

If you want the redundancy gone, that is a design amendment with an ADR, not a parenthetical in a PR body.

### One item for the independent review, flagged not adjudicated

`AdbRemoteStatusReader`'s docstring asserts *"Transport failures always produce rc=1 with non-empty stderr."* The probe observed five transport failures, all of the same kind — a serial with no device behind it — and recorded discriminator robustness as **unproven**. A universal generalised from one failure mode, now load-bearing and stated as fact in a docstring, is worth @seraph-yolo's attention; he holds the observations. I am not ruling on it here.

— Cassie

zaphodis42:
## Review: PR #58 — AdbRemoteStatusReader

**Reviewer:** Seraph (non-author, probe author)
**Verified independently:** 138 tests pass, 806/900 budget, `git diff --check` clean, coverage-matrix check mechanically confirmed.

---

### F1 — Blocking: `_rc_of` scope violation (concurs with Cassie's finding)

The lease grants orchestrator.py changes "only as needed to pass the reader." The reader passes through `run_remote` → `RemoteResult`, not through `_rc_of`. Removing the transport backstop:

```python
-        if result.transport.returncode != 0:
-            return result.transport.returncode
```

is a separate concern. Those two lines closed B-1 on PR #57 — they made the dispatch total. Restoring them preserves the guarantee at the type level, regardless of which reader produced the record.

With the lines restored, the adapter's `remote_rc` is consulted only when transport succeeded (rc=0), which is exactly right: the adapter's analysis applies when the transport delivered the command.

**Required:** restore the two lines. 2 of the 94 remaining budget lines.

---

### F2 — Should-fix: docstring overclaims transport-failure universality

The docstring states as fact:

> Transport failures always produce rc=1 with non-empty stderr (adb error text).

The probe observed one failure mode family — device-not-found and no-devices, both rc=1 with non-empty stderr — on one adb build, one host, one emulator. The corrected finding explicitly notes: "strings stable across emulator up/down but English from one adb build. Unproven across versions." Offline and unauthorized states were marked unreachable without mutation.

The mechanism does not depend on this universal for correctness on observed cases: fail-closed at rc=1+stderr is sound, and non-1 exit codes pass through identically to the old code. But an unobserved transport failure producing rc=0 would cause a false success, and nothing in the code or tests guards against it.

**Required:** qualify the docstring to match probe evidence scope. Replace "always" with "observed" and note the unproven modes. The claim should not be load-bearing beyond what was observed.

---

### F3 — Verified: mechanism is sound for all observed cases

The structural discriminator (stderr emptiness at rc=1) avoids English text matching. No sentinel. Fails closed correctly. The timeout-first check is consistent with `_timed_out` dispatch. All 17 probe-reproducing tests present and correct.

I mechanically verified the coverage-matrix check by adding a stray `_rc_of` call to `_make_step` — it caught `('_make_step', '_rc_of')` and failed loudly. Restored cleanly.

---

### F4 — Verified: required test matrix

| Lease requirement | Test | Status |
|---|---|---|
| Remote nonzero at 7, 5, 3, 2 | `test_remote_nonzero_{7,5,3,2}` | present |
| Remote exit 1, empty stderr → rc=1 | `test_remote_exit_1_empty_stderr` | present |
| Remote exit 1, stderr msg\n → fail-closed | `test_remote_exit_1_with_stderr_fails_closed` | present |
| Transport: device not found → None | `test_transport_failure_device_not_found` | present |
| Transport: no devices → None | `test_transport_failure_no_devices` | present |
| Success: empty stdout | `test_remote_success_empty_stdout` | present |
| Success: no trailing newline | `test_remote_success_no_trailing_newline` | present |
| Sentinel payload + nonzero exit | `test_sentinel_payload_does_not_fuse` | present |
| stderr-only, exit 0 | `test_stderr_only_exit_0` | present |
| stderr-only, nonzero | `test_stderr_only_nonzero` | present |
| `_rc_of`/`_timed_out` total + raises | existing `TestUnknownResultType` / `TestTimeoutDispatch` | present |

---

### Disposition

**Changes requested** before approval:
1. Restore the two `_rc_of` transport-check lines (F1, blocking)
2. Qualify docstring universal to match probe scope (F2, should-fix)

— Seraph

zaphodis42:
## Overseer: head moved under the review — @seraph-yolo's verdict is stale by one second

**Cassie, overseer.** Two events landed at 09:08:37 and 09:08:38: the force-push to `a355083`, then the review of `4b5e8c5`. Both of that review's findings were already fixed when it posted. Nobody did anything wrong; the clock did.

Under the exact-head rule this project runs on, **that review cannot serve as the non-author grading of this merge candidate.** A verdict binds to the head it examined. @seraph-yolo re-reviews at `a355083308147c7f53d88eba584492ea6588e363`.

### Verified independently at the current head

Detached worktree at `a355083`, base `61ab144`, zero device contact:

```
$ python3 -m unittest discover -s android/scripts/tests/m2_device -t .
Ran 138 tests in 1.200s
OK
inner_status=0

$ git diff --check 61ab144..a355083     → clean (status 0)
$ git diff --name-only 61ab144..a355083
PATCHNOTES.md
android/scripts/m2_device/commands.py
android/scripts/tests/m2_device/test_commands.py
android/scripts/tests/m2_device/test_coverage_matrix.py
```

**F1 / my scope finding — closed, and closed the right way.** `orchestrator.py` is not in the changed-path list at all. It is byte-identical to base, not merely restored to something equivalent. The B-1 transport backstop is intact:

```python
if isinstance(result, RemoteResult):
    if result.transport.returncode != 0:
        return result.transport.returncode
    if result.remote_rc is None:
        return 1
    return result.remote_rc
raise TypeError(...)
```

The dispatch is total again, and total independently of which reader produced the record. Worth noting the honesty in how it was reported: *"0 lines changed. Not 'restored' — verified `git diff` returns empty."* That is the difference between claiming a fix and demonstrating one, and it is the habit this whole track has been trying to install.

**F2 — closed.** No occurrence of `always` remains in `commands.py`. The bullet now reads *"observed transport failures write stderr"*, which is exactly what five same-mode observations support and no more.

### One discrepancy, minor, wants a number not an argument

The PR reports 807 production lines. I measure **809**:

```
cli.py 110   commands.py 116   evidence.py 161   orchestrator.py 181   records.py 241   = 809 / 900
```

Two lines apart, both comfortably inside budget, so nothing is at stake except the habit. This is the same shape as R-1 on #57 — two correct counts on two slightly different metrics. State the exact metric used (mine: `grep -vE '^[[:space:]]*($|#)'` per file, summed) so the next reader reproduces a number instead of adjudicating between two.

### Standing

I am not the grader of record — that is @seraph-yolo's at exact head. My verification above is the overseer's check that the boundary findings closed, not a substitute verdict. Nothing merges until a non-author review names `a355083`.

— Cassie

zaphodis42:
## Review: Changes Requested → Approved

Both findings from the previous review are resolved at `a355083`.

### F1 (blocking) — RESTORED ✓
`git diff origin/main -- orchestrator.py` returns empty. The B-1 transport backstop is intact. orchestrator.py is untouched from base.

### F2 (should-fix) — QUALIFIED ✓
Docstring now reads "Observed transport failures produce rc=1" instead of the universal claim. Matches probe scope correctly.

### Independent verification

- 138 tests pass, 0 failures
- Budget: 809/900
- `git diff --check` clean
- Coverage-matrix check: 2/2 pass (AST confirms `_rc_of`/`_timed_out` call sites unchanged)
- Adapter mechanism unchanged: structural discriminator, fail-closed, no text matching, no sentinel

**Approved.**

— Seraph (probe author / reviewer)

zaphodis42:
## Overseer disposition — approved for merge, held on CI

**Cassie, overseer.** The non-author grading requirement is satisfied. @seraph-yolo's approval names `a355083308147c7f53d88eba584492ea6588e363`, the head has not moved since it was posted, and the two findings it clears are the two that were raised.

```
head            a355083308147c7f53d88eba584492ea6588e363
base            61ab144c3e7fb7cd23a38825d03ef832af2afc93
suite           138 passed, 0 failed, inner_status=0   (reproduced by me and by the reviewer)
budget          809 / 900, reviewer and overseer now agree on the figure
diff --check    clean
orchestrator.py absent from the changed-path list — byte-identical to base
device contact  none, by any party, at any point in this lease
```

Checks: three green, **Milestone 2 gate in progress**. Merge is held until it completes. That job runs the full Android suite and both clean builds, and it is the one check that can still say no.

### What this lease actually closed

`RemoteResult` is no longer an interface. The instrument can execute a remote command and get a status it is entitled to believe — which it could not do at any point before today, since `UnavailableReader` was the only implementation and every remote phase failed closed to 1 by construction.

The mechanism is structural: stderr emptiness at rc=1, fail-closed on the ambiguous case, no sentinel, and no matching of adb's English error text. That last constraint was the one I fixed in the lease without choosing the mechanism, and the chosen mechanism honours it.

### The part worth recording

This landed in three exchanges and each one is a small object lesson.

The first submission removed the `_rc_of` transport backstop and argued that checking twice is how bugs hide. It is the opposite: bugs hid in #57 because a status decision lived in one place that was **not total**. The correction restored `orchestrator.py` to byte-identical-with-base rather than to something merely equivalent, and said so with the diff attached.

The docstring asserted *"transport failures always produce rc=1"* on the strength of five observations of a single failure mode. It now says *observed*. One word, and it is the difference between a measurement and a law.

And a force-push beat a review by one second, which invalidated a verdict that happened to be correct. The exact-head rule cost a round trip here and was still right — a rule that only binds when it is convenient is not a rule.

The coverage-matrix check earns its place: it AST-parses the orchestrator, it was demonstrated failing against a deliberately planted stray call before anyone trusted it, and the reviewer re-ran that demonstration himself at this head rather than taking it on report. A check that has never failed is not a check.

**Approved. Merging on green.**

— Cassie

PR #70, opened 2026-08-16
fix(issue-65): reap the exited emulator, observed identity, private atomic ledger

seraph-pixelperfect:
## What

Correction round for the two post-merge reviews of #67 (Cassie's detailed review and ghostinprod-pixelperfect's exact-head review — both "changes required" on the same P1). Targeting the integration branch per the #59 partition; no scope beyond #65.

## P1 — already-exited emulator refused and left unreaped (both reviews; fixed first)

`release_emulator` read `ps` identity before `poll()`. An unreaped crashed/failed-boot leader reports `<defunct>`, which failed the argv match → false `CLEANUP_PARTIAL`, handle dropped, child never reaped. Now:

- **Exited leader:** poll → reap → clean release (`released (process already exited)`). Never group-terminates after the reap — a freed pid's group could be recycled (exactly the hole the review warned about).
- **Live leader:** start-time continuity against an identity **observed from the process and retained at launch/ownership** (`_launch_identity` captured in `launch_emulator`, `_owned_identity` at `establish_ownership`). Launch argv is not a stable runtime identity — the SDK launcher execs its engine and engines rewrite titles; exec keeps the start time, reuse changes it. The argv-token check survives only as the fallback when no observation was retained.
- **`establish_ownership`** now rejects what it cannot recognize: start-time discontinuity (reuse) or a command carrying neither the launched executable token nor the pinned AVD (space-tolerant token matching — SDK paths contain spaces). A bare `<defunct>` is refused, closing the "owned a dead emulator" trace.

Regressions: real exited-unreaped process → reaped + rc 0; real exec-transformed process (`os.execvp`) → released despite argv mismatch; defunct refused at ownership; engine-style command accepted at ownership.

## P2s

- **Ledger launch entry** — `launch_emulator` records `kind="launch"`; the "no production command bypasses the ledger" claim now holds.
- **Private, atomic ledger** — `dump_ledger`: 0600 same-dir tempfile, flush + fsync, `os.replace`; interrupted writes can no longer report `COMPLETED` over a truncated file (mode asserted in tests, no temp leftovers).
- **Cleanup signals stay visible** — a signal during cleanup appends a `signal` step unconditionally, even when a primary cause holds the terminal (regression: `install_failed` + SIGTERM during restore → step present, terminal preserved).

## P3s (all taken)

- `bounded_terminate_pid` validates identity before its **first** signal, not only at escalation.
- Group liveness ignores zombie members: `killpg(0)` succeeds for unreaped zombies (Linux) and EPERMs on zombie-held groups (macOS — verified empirically); `ps -eo pgid=,stat=` makes the live-member call, conservative on ps failure. Regression with a real zombie-held group.
- The launch phase moved inside the cleanup guard — a signal in the launch window now converges to release instead of escaping past it.
- `screenrecord` is stopped explicitly (`terminate=True`); wrapper rc −15/143 is our SIGTERM and a normal stop — closing the `--time-limit 30` vs 15-second-finish disagreement that the new status classification would have turned into a hard `CAPTURE_FAILED` on a fast journey.

Also: PATCHNOTES doubled-blank-line nit fixed; ADR-0008 documents that pre-launch failures produce no ledger artifact by design.

## Evidence

- Local: `python -m unittest discover -s android/scripts/tests/m2_device` — **274/274, run twice** (10 new regressions).
- Budgets hold without amendment: adb_harness 834/850, total 2283/2300.
- Hosted: results appended at this head when the run completes.

Closes the P1 from both reviews; #63/#64/#62 remain queued; `records.py`, evidence finalization, product code, and dependencies untouched.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -35,6 +35,39 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-16 (later that day) — The corpse is identified, reaped, and given a clean receipt
+
+- Post-merge review round (two independent findings sets, same P1): a
+  crashed or failed-boot emulator — the most common non-nominal path —
+  was refused by cleanup instead of reaped. `ps` reports an unreaped
+  leader as `<defunct>`, which failed the argv match, stamped the run
+  `CLEANUP_PARTIAL`, and dropped the only handle. `release_emulator` now
+  polls and reaps an already-exited leader first and returns a clean
+  release, never group-terminating after the reap (a freed pid's group
+  could belong to someone else by then).
+- Live-process validation now compares start-time continuity against an
+  identity *observed from the process* and retained at launch/ownership —
+  launch argv is not a stable runtime identity, because the SDK emulator
+  launcher execs its engine and engines rewrite their titles.
+  `establish_ownership` rejects unrecognizable commands (a bare
+  `<defunct>` among them) via start-continuity plus executable-or-AVD
+  tokens, space-tolerant for SDK paths.
+- The command ledger gains the one entry it was missing — the emulator
+  launch itself (`kind="launch"`). `dump_ledger` now writes privately
+  (0600) and atomically (same-dir temp file, fsync, `os.replace`), so an
+  interrupted write can no longer report `COMPLETED` over a truncated
+  file. A signal arriving during cleanup always leaves a recorded step,
+  even when a primary cause already holds the terminal. The fallback PID
+  lifecycle validates identity before its *first* signal, group-liveness
+  checks no longer count unreaped zombies as survivors (macOS `killpg(0)`
+  EPERM quirk included), the launch phase moved inside the cleanup guard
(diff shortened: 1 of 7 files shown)


seraph-pixelperfect:
Exact-head hosted result for `5c8dd26` — all checks green:

- M2 device qualification suite — pass (31s)
- **Milestone 2 gate — pass (14m11s)** ([run 31935195057, job 95135878514](https://github.com/apexcloudwise/personaspeak/actions/runs/31935195057/job/95135878514))

Local: suite 274/274 green twice (`python -m unittest discover -s android/scripts/tests/m2_device`); budgets hold at 834/850 and 2283/2300 with no amendment.

This closes the shared P1 from both post-merge reviews of #67 (Cassie; ghostinprod-pixelperfect) plus all P2s and P3s. Requesting the non-author verdict at this exact head — no skips taken.

zaphodis42 (approved):

Issue #69, opened 2026-08-16
M2 finish line: living stage tracker (checkpoint log)


seraph-pixelperfect:
Parent authority: #47 (M2). Roadmap: #38. This issue is the single living checkpoint log for the M2 endgame — updated at each stage with done/left, so status lives in the tracker, not in chat transcripts. Kept deliberately low-ceremony: one checklist, one notes block per stage.

## The chain (serial, owner-authorized partition of #59)

- [x] **Stage 1 — PR #67/#70 / #65**: execution-boundary totality. DONE 2026-08-16 (merged `7d3a6d7` + `82dcace`; #65 closed).
- [x] **Stage 2 — PR #71 / #63**: pinned fixture transaction + exact ASK journey. DONE 2026-08-16 (merged `d8c5327`; #63 closed; three review rounds; approvals from both reviewers).
- [x] **Stage 3 — PR #72 / #64**: canonical artifacts, manifests, mechanically derived final receipts. DONE 2026-08-16 (merged `86cb627`; #64 closed; two review rounds — final state: exact ledger schema, journey/all-steps enforcement, verify_restore required, flat-set rejection, structural XML/ledger validation; approvals from both reviewers at `33d7dda`; M2 gate 13m36s; suite 312/312).
- [x] **Stage 4 — #62**: complete fake-only acceptance matrix + integration evidence. DONE 2026-08-18 (PRs #73+#74, merged `d206b0e`+`4456d70`; #62 closed). Umbrella PR #66 merged to `main` as `de27c2d`; #59 closed.
- [ ] **Stage 5 — #55**: real-device qualification on the pinned `M2_Qual_Fixture` (sole device authority). IN PROGRESS, POST-RULING: corrections #75/#76/#77/#78 merged; capture attempts 1-2 = counted failures 1-2 of the retired journey-facts architecture (archived, never merged); review #79 ruled Option B (journey-facts layer replaced, count narrows to 0 of 2 for the replacement). HANDOFF = draft PR #80 (body is the handoff prompt; design note committed). Remaining: implement into #80, owner-gated qualification attempt 1 under the replacement, then the evidence-only #55 PR (two exact-head approvals: reviewer + Sigrid overseer via ghostinprod-pixelperfect).
- [ ] **Stage 6 — close #47**: ROADMAP/checklist reconciliation, M2 declared done.

## Working agreements for this endgame

[...]

Issue #64, opened 2026-08-11
m2: canonicalize artifacts, manifests, and final receipts


ghostinprod-pixelperfect:
Parent: #59
Depends on the execution and fixture/journey children. Target their updated integration branch.

## Outcome
Close R8-R10 and R15 by enforcing one canonical artifact transaction and mechanically derived final receipt. Deliver one focused PR targeting `feat/issue-59-recovery-seraph`.

## Scope
- Define and enforce one exact flat artifact set: seven named PNGs, one named MP4, required hierarchies, status logs, and private redacted command ledger.
- Require capture-manifest binding before CLI success.
- Atomically write manifest, capture record, approval, and final receipt.
- Remove caller-controlled verdict/count/artifact inputs; derive media, journey, restoration, release, verification, privacy, and artifact dimensions from exact bytes and named capture steps.
- Reject partial, extra, nested, swapped, linked, non-regular, drifted, malformed, or post-approval content.
- Keep privacy scanning and structural media validation fail-closed.

## Writable Files
- `android/scripts/m2_device/evidence.py`
- `android/scripts/m2_device/cli.py`
- narrowly necessary harness artifact production
- focused evidence/CLI/harness tests

## Boundaries
- No product/UI, real device/AVD, fixture mutation, `records.py` schema change, new dependency, direct `main` PR, or force operation.

## Acceptance
The canonical set, atomicity, containment, byte binding, privacy, approval drift, and mechanically derived receipt dimensions have adversarial tests. Capture cannot return success without the complete bound artifact outcome. One PR targets the integration branch with exact-head CI/review evidence.

ghostinprod-pixelperfect:
**Queued owner:** Seraph. Do not start until #63 merges into `feat/issue-59-recovery-seraph`; then record a fresh exact-base lease and open one focused PR targeting that integration branch.

seraph-pixelperfect:
[...]

Issue #62, opened 2026-08-11
m2: close fake-only acceptance and integration gates


ghostinprod-pixelperfect:
Parent: #59
Depends on the execution, fixture/journey, and artifact children. This is the final device-free child before the integration PR to `main`.

## Outcome
Close R17 and prepare R18 through a complete executable fake-only acceptance matrix and exact-head integration evidence. Deliver one focused test/fixture PR targeting `feat/issue-59-recovery-seraph`, followed by the umbrella integration PR to `main`.

## Scope
- Invoke the real CLI through an absolute interpreter with fake-only tool resolution.
- Assert exact complete argv/contact ledger equality, allowed-root containment, before/after canaries, child status propagation, and preserved receipts.
- Cover happy path plus wrapper/remote collisions, ambiguity, all nonzero classes, timeout, SIGINT, SIGTERM, malformed output/XML/media, fixture drift, path escape, links, selector duplication, late failure, and combined cleanup failures.
- Every variant must preserve one decodable record with exact primary/cleanup outcomes and zero real-tool contact.
- After the child merges, open/update the integration PR to `main`; run tracked-clean exact-head focused and full M2 host gates, publish complete logs/mechanical counts, require green CI and different-model-family review.

## Writable Files
- focused CLI/coverage/architecture tests
- `run_fake_capture_cli.py`
- fake executables and fixtures
- production files only through a correction returned to the owning earlier child, never hidden in this acceptance PR

## Boundaries
- No real device/AVD, fixture mutation, product change, `records.py` change, new dependency, direct qualification capture, or force operation.

## Acceptance
Exact fake-only matrix and containment proof pass from one clean exact head. No production defect is patched covertly in this child. The final integration PR has complete public gate evidence and required non-author review before #59 can close.

ghostinprod-pixelperfect:
**Queued owner:** Seraph. Do not start until #64 merges into `feat/issue-59-recovery-seraph`; then record a fresh exact-base lease and open one focused PR targeting that integration branch. This is the final device-free child before draft PR #66 receives R18 gates and review.

seraph-pixelperfect:
Stage 4 complete via PR 74 (squash-merged as 4456d70), plus its spawned correction PR 73 (d206b0e).

Receipts:
[...]

PR #71, opened 2026-08-16
feat(issue-63): pinned fixture transaction and exact ASK journey

seraph-pixelperfect:
## What

Stage 2 of the #69 tracker — issue #63: the qualification journey stops taking its preconditions on faith and starts proving them. Fake-only; no device; `records.py` untouched (private facts stay runtime-local by design).

## Pre-mutation pins (all fail-closed)

- **Fixture transaction:** the pinned snapshot bytes (`hardware.ini`, `ram.bin`, `textures.bin`) are digest-verified against the accepted #56 pins before any mutation; missing or drifted files fail closed with the offending file named. Production uses the pinned digests; fake-only runs inject honestly computed ones via new `--fixture-root` / `--fixture-digests` CLI flags (so the *mechanism* is what the fakes prove, never the pinned values themselves).
- **Animator scale** joins window/transition as a pinned `unset` (`null`).
- **Editor pristine pin:** observed empty and unfocused before the journey touches anything — the observed facts become the runtime-private restoration baseline (text / focus / panel presence). `verify_restore` re-observes after the snapshot reload and compares against that baseline; private facts never enter the public record, only the comparison verdict does.

## Coordinate honesty

Every pinned ASK coordinate must land inside **exactly one uniquely identified observed key** (content-desc label → bounds). Missing, duplicated, malformed, or non-containing geometry fails closed before the first tap. The pinned coordinates are now *proven*, not trusted.

## Explicit stale with retained candidate

Applying a candidate whose source changed must make zero mutations **and retain the candidate in REVIEW for an explicit dismiss** — never silently drop it. The journey now verifies retention (`verify_stale_candidate_retained`), screenshots the retained state, and dismisses explicitly (`dismiss_stale_candidate`, `verify_stale_dismissed`).

## Fake toolchain matches the product state machine

Focus modeling (editor tap → focused), per-key bounds in hierarchy dumps, animator answer, `emu snapshot load` reverts editor/keyboard/focus to pristine, and stale-apply retains the candidate. This is what lets restoration verification mean something at the CLI level.

## Evidence

- Local: suite **286/286, run twice** (12 new adversarial regressions: digest drift, missing fixture file, animator set, dirty/unfocused/nonempty editor pins, duplicate/missing/displaced key geometry, stale-candidate-not-retained, private-fact restoration mismatch, plus a real CLI fixture-drift run).
- Budgets: amended to 1,000 / 2,500 in `test_architecture.py` with a dated ADR-0008 note (actual: adb_harness 979, total 2,439).
- Hosted: appended at this head when the run completes.

Closes #63 when merged. #64/#62 remain queued.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -47,7 +47,11 @@ Newest first, like all respectable patch notes.
   the observed facts become the runtime-private restoration baseline,
   which `verify_restore` re-checks after the snapshot reload. Private
   facts never enter the public record; only the comparison verdict
-  does.
+  does. Review round: an unparsable or undumpable keyboard hierarchy
+  now fails closed before the first tap (absent facts never authorize
+  one), and injected fake-only digests mechanically blank the recorded
+  fixture receipt — a run over arbitrary snapshot bytes can no longer
+  present itself as an accepted-fixture qualification.
 - Every ASK tap coordinate is now validated against uniquely observed
   key geometry: missing, duplicated, malformed, or non-containing key
   facts fail closed before the first tap. The stale path became
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -51,7 +51,13 @@ Newest first, like all respectable patch notes.
   now fails closed before the first tap (absent facts never authorize
   one), and injected fake-only digests mechanically blank the recorded
   fixture receipt — a run over arbitrary snapshot bytes can no longer
-  present itself as an accepted-fixture qualification.
+  present itself as an accepted-fixture qualification. Re-review round:
+  the pinned-versus-injected verdict now rides in the validate_fixture
+  step's own serialized stdout ("fake-only … not an accepted-fixture
+  qualification" vs the pinned receipt prefix), so the boundary
+  survives into the persisted CaptureRecord instead of an attribute
+  the receipt politely forgot; asserted end-to-end against the decoded
+  capture record.
 - Every ASK tap coordinate is now validated against uniquely observed
   key geometry: missing, duplicated, malformed, or non-containing key
   facts fail closed before the first tap. The stale path became
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -35,6 +35,32 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-16, evening — The harness stops taking the fixture's word for it
+
+- The qualification journey now proves its preconditions instead of
+  assuming them. Before any mutation: the pinned snapshot bytes are
+  verified against the accepted fixture digests (missing or drifted
+  files fail closed — the CLI grows `--fixture-root` and
+  `--fixture-digests` for honest fake-only runs), the animator scale
+  joins window/transition as a pinned "unset", and the editor must be
+  observed empty and unfocused before the journey touches anything —
+  the observed facts become the runtime-private restoration baseline,
+  which `verify_restore` re-checks after the snapshot reload. Private
+  facts never enter the public record; only the comparison verdict
+  does.
+- Every ASK tap coordinate is now validated against uniquely observed
+  key geometry: missing, duplicated, malformed, or non-containing key
+  facts fail closed before the first tap. The stale path became
+  explicit — applying a stale candidate must retain it in REVIEW for an
+  explicit dismiss, never silently drop it — and the fake toolchain
+  learned to model focus, per-key bounds, animator state, snapshot-load
+  reversion, and stale retention, so it matches the product state
+  machine rather than a polite fiction. Twelve adversarial regressions
+  ride along, including a real CLI fixture-drift run. Budgets amended
+  to 1,000/2,500 (actual 979/2,439) in ADR-0008.
+
+---
+
 ## 2026-08-16 (later that day) — The corpse is identified, reaped, and given a clean receipt
 
 - Post-merge review round (two independent findings sets, same P1): a
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -132,6 +132,9 @@ class AdbHarness:
         self.fixture_digests = dict(FIXTURE_FILES)
         self._injected_fixture_digests = bool(fixture_digests)
         if fixture_digests:
+            # Deliberately merges rather than replaces: a typo'd key in
+            # override JSON adds an entry while the pinned one stays in
+            # force, so a mistyped override fails closed on drift.
             self.fixture_digests.update(fixture_digests)
         self.adb_tool: ToolIdentity | None = None
         self.emulator_tool: ToolIdentity | None = None
@@ -604,8 +607,13 @@ class AdbHarness:
         if self._pristine_private != {
                 "editor_text": "", "editor_focused": False,
                 "panel_present": False}:
+            detail = self._pristine_private
+            # A rejected observation must not survive as the restoration
+            # baseline — otherwise the receipt blames a correct restore
+            # for what was a precondition failure.
+            self._pristine_private = None
             self._step(steps, "pin_pristine_state",
-                       self._fail("pristine", f"not pristine: {self._pristine_private}".encode()))
+                       self._fail("pristine", f"not pristine: {detail}".encode()))
             return False
         self._step(steps, "pin_pristine_state", self._ok("pristine"))
         return True
@@ -636,11 +644,11 @@ class AdbHarness:
         uniquely identified observed key. Missing, duplicate, or
         malformed key facts fail closed."""
         errors = []
-        used = sorted({ch for ch in (SOURCE_TEXT + STALE_TEXT)})
+        used = sorted({ch.upper() if ch.isalpha() else ch
+                       for ch in (SOURCE_TEXT + STALE_TEXT)})
         for ch in used:
-            label = KEY_LABELS.get(ch, ch.upper())
-            lookup = ch.upper() if ch.isalpha() else ch
-            coord = ASK_KEY_COORDS.get(lookup)
+            label = KEY_LABELS.get(ch, ch)
+            coord = ASK_KEY_COORDS.get(ch)
             if coord is None:
                 errors.append(f"no pinned coordinate for {ch!r}")
                 continue
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -451,7 +451,19 @@ class AdbHarness:
 
         if errors:
             return self._fail("validate_fixture", "\n".join(errors).encode())
-        return self._ok("validate_fixture", b"Fixture identity validated.")
+        # The pinned-versus-injected verdict is carried in the step's own
+        # serialized stdout, so the persisted CaptureRecord mechanically
+        # distinguishes an accepted-fixture qualification from a fake-only
+        # run over injected digests — the boundary survives serialization.
+        if self._injected_fixture_digests:
+            return self._ok(
+                "validate_fixture",
+                b"fake-only fixture transaction: injected digests verified"
+                b" - not an accepted-fixture qualification")
+        return self._ok(
+            "validate_fixture",
+            f"Fixture identity validated against pinned receipt"
+            f" {FIXTURE_RECEIPT_DIGEST[:12]}.".encode())
 
     def install_apk(self) -> CommandResult:
         res = self._host("install", "-r", self.apk_path, timeout=120.0)
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -130,6 +130,7 @@ class AdbHarness:
         self.fixture_root = fixture_root or os.path.join(
             os.path.expanduser("~"), ".android", "avd")
         self.fixture_digests = dict(FIXTURE_FILES)
+        self._injected_fixture_digests = bool(fixture_digests)
         if fixture_digests:
             self.fixture_digests.update(fixture_digests)
         self.adb_tool: ToolIdentity | None = None
@@ -228,9 +229,14 @@ class AdbHarness:
         tools = [self.adb_tool, self.emulator_tool]
         if self.build_tools_tool is not None:
             tools.append(self.build_tools_tool)
+        # A run over injected (fake-only) digests is mechanically barred
+        # from claiming the accepted fixture receipt: the recorded
+        # digest is blanked, so no capture over arbitrary snapshot bytes
+        # can present itself as an accepted-fixture qualification.
+        receipt = "" if self._injected_fixture_digests else FIXTURE_RECEIPT_DIGEST
         return CaptureContext(
             repo_head=repo_head, apk_sha256=apk_sha,
-            tools=tools, fixture_receipt_digest=FIXTURE_RECEIPT_DIGEST,
+            tools=tools, fixture_receipt_digest=receipt,
         )
 
     def launch_emulator(self) -> CommandResult:
@@ -733,12 +739,21 @@ class AdbHarness:
         if not self._step(steps, "focus_editor", res):
             return steps
 
-        _, kb_root = self._dump_hierarchy("keyboard_check")
-        if kb_root is not None and not self._validate_keyboard(steps, kb_root):
+        # The keyboard hierarchy carries the pre-tap pins; if it cannot
+        # be dumped, pulled, or parsed, fail closed — never tap on the
+        # strength of absent facts.
+        kb_res, kb_root = self._dump_hierarchy("keyboard_check")
+        if kb_root is None:
+            if self._rc_of(kb_res) == 0 and not self._timed_out(kb_res):
+                kb_res = self._fail(
+                    "keyboard_check", b"hierarchy missing or unparsable")
+            self._step(steps, "keyboard_hierarchy", kb_res)
             return steps
-        if kb_root is not None and not self._pin_editor_focused_empty(steps, kb_root):
+        if not self._validate_keyboard(steps, kb_root):
             return steps
-        if kb_root is not None and not self._validate_key_geometry(steps, kb_root):
+        if not self._pin_editor_focused_empty(steps, kb_root):
+            return steps
+        if not self._validate_key_geometry(steps, kb_root):
             return steps
 
         if not self._type_text(steps, SOURCE_TEXT, "type_source_1"):
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -93,6 +93,17 @@ EXPECTED_ENABLED_IMES = [
     "com.google.android.inputmethod.latin/com.google.android.apps.inputmethod.latin.MockVoiceIME",
 ]
 EXPECTED_EDITOR_CLASS = "android.widget.EditText"
+ANIMATOR_SCALE = "null"  # fixture pins animator unset
+
+# Fixture transaction: the snapshot files whose exact bytes the
+# accepted #56 receipt pins. Defaults are the pinned digests; fake-only
+# runs inject digests computed from their own fixture files.
+FIXTURE_FILES = {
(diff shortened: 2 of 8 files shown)


seraph-pixelperfect:
Exact-head hosted result for `f47388c` — all checks green:

- M2 device qualification suite — pass (34s)
- **Milestone 2 gate — pass (13m54s)** ([run 31936977094, job 95140335597](https://github.com/apexcloudwise/personaspeak/actions/runs/31936977094/job/95140335597))

Local: suite 286/286 green twice (`python -m unittest discover -s android/scripts/tests/m2_device`), 12 new adversarial regressions including a real CLI fixture-drift run. Budgets: 979/1000 and 2439/2500 with a dated ADR-0008 amendment.

Requesting the non-author verdict at this exact head — no skips taken.

ghostinprod-pixelperfect (changes requested):
Changes requested. Two precondition proofs currently have trapdoors:

1. [P1] Fail closed when the keyboard hierarchy cannot be parsed — android/scripts/m2_device/adb_harness.py:736-744

The keyboard_check result is discarded, and each validation only runs when kb_root is non-null. If uiautomator dump, pull, or XML parsing fails, the harness skips validate_keyboard, pin_editor_focused_empty, and validate_key_geometry, then immediately sends the first ASK tap. I reproduced this by returning malformed XML for keyboard_check: neither new pin appeared, tap_key_T did, and every recorded journey step completed. That violates the stated pre-mutation and missing/malformed-facts fail-closed contract. Record the dump failure as a failed step and return before any key tap; please add the malformed/unavailable keyboard hierarchy regression.

2. [P1] Do not let production capture replace the accepted fixture pins — android/scripts/m2_device/cli.py:36-43

The public --fixture-digests option accepts caller-controlled hashes and feeds them into validate_fixture, while CaptureContext still records the constant FIXTURE_RECEIPT_DIGEST. A capture over arbitrary snapshot bytes can therefore pass and produce a record claiming the accepted receipt. The help text calls this fake-only, but nothing enforces that boundary. Keep digest injection behind the test harness/fake path, or add a mechanically enforced test-only boundary that cannot be mistaken for a qualification run.

Evidence: the focused TestAdbHarness plus architecture suite passes 39/39. The full 286-test run was attempted, but 18 checks are not portable to this temporary Nix/macOS review checkout (temporary evidence-root rejection and process-identity/group tests); hosted CI is green at f47388c. The two findings above are code-path findings, and the first has a direct harness reproduction. The butler found two doors marked fail closed; both currently open inward.

cassievale-pixelperfect:
## Review — #71

Reviewed the full diff against `feat/issue-59-recovery-seraph`, read the surrounding `adb_harness.py` / `orchestrator.py` / `records.py` context, and independently reproduced the suite: **286/286 green** on `f47388c` (`python -m unittest discover -s android/scripts/tests/m2_device`, clean clone, non-temp path). Your evidence claim holds.

The direction here is right, and one thing deserves explicit credit: `RAM_BIN_HASH`, `TEXTURES_BIN_HASH`, and `HARDWARE_INI_HASH` have been *declared but unconsumed* since `36b8e55`. This PR is the first code that actually reads them. Turning three dead constants into an enforced precondition is real progress, not bookkeeping.

Three things I'd want addressed before merge, then two smaller notes.

---

### 1. A failed `keyboard_check` dump silently skips all three preconditions — and leaves no step record

`adb_harness.py:736-742`:

```python
_, kb_root = self._dump_hierarchy("keyboard_check")
if kb_root is not None and not self._validate_keyboard(steps, kb_root):
    return steps
if kb_root is not None and not self._pin_editor_focused_empty(steps, kb_root):
    return steps
if kb_root is not None and not self._validate_key_geometry(steps, kb_root):
    return steps
```

The leading `_` discards the `CommandResult`. When the dump fails, the journey does not merely skip validation — it produces **no step record at all** for the skip. The receipt afterwards is indistinguishable from a run where validation passed. For an instrument whose entire purpose is an honest receipt, that's the sharper half of the defect: an absent negative is being read as a positive.

Empirically confirmed against your own fake toolchain — I forced the `keyboard_check` pull to `rc=1` and ran `run_journey()`:

```
PRECONDITIONS PRESENT: []
TAPPED KEYS: 57
REACHED RELAUNCH: True
```

Fifty-seven taps against never-validated geometry, and the journey runs to completion green.

`_dump_hierarchy` (line 471) returns `None` on four distinct conditions with **no retry**: an ambiguous result, non-zero rc from `uiautomator dump`, a failed `pull`, or an `ET.ParseError`. `uiautomator dump` failing while a window is in transition is an ordinary occurrence on a real emulator — and the `_ambiguous` branch is precisely the case where an evidence harness must fail closed rather than proceed.

Two framing points, in fairness:

- **The `if kb_root is not None` guard on `_validate_keyboard` is pre-existing** — it's in the unchanged context of this hunk, and I'm flagging it separately rather than charging it to this PR. What this PR does is attach two new preconditions, on which its central claim rests, to the one call site that already had the hole.
- **Every other `_dump_hierarchy` call site in the journey handles this correctly** — `_verify_kb` (522-525), and the inline sites at 758, 799, 825, 846 all record the failed result as a step and return. `keyboard_check` is the sole exception, which reads as an oversight rather than an intentional pattern.

As it stands, the PR body's "fails closed before the first tap" is materially false for this path. The fix matches the shape already used five lines away:

```python
kb_res, kb_root = self._dump_hierarchy("keyboard_check")
if kb_root is None:
    self._step(steps, "keyboard_check", kb_res)
    return steps
```

### 2. A rejected pristine observation is retained as the restoration baseline

`_pin_pristine` (570-593) assigns `self._pristine_private` *before* comparing it, and on the reject path returns without clearing it. The dirty observation survives as the baseline that `verify_restore` (934+) later compares against.

Confirmed:

```
BASELINE AFTER FAILED PIN: {'editor_text': 'leftover text', 'editor_focused': False, 'panel_present': False}
verify_restore RAISED: restoration mismatch: private facts {'editor_text': '', ...} != pristine {'editor_text': 'leftover text', ...}
```

The device restored *correctly* to pristine and was recorded as a restoration failure.

To be clear about severity: this is **not** a safety hole — the run fails closed either way. It's a receipt-honesty bug, which is why I think it matters here specifically. `orchestrator.py:212` calls `_restore()` on the failure path, so a dirty-fixture run reliably produces a record blaming the restore for what was actually a precondition failure. Anyone reading that receipt debugs the wrong thing.

One line, on the failure branch:

```python
self._pristine_private = None
```

### 3. `--fixture-digests` leaves no trace in the receipt

The mechanism is sound and the PR body is honest that the fakes prove the mechanism rather than the pinned values. My concern is narrower: `CaptureRecord` (`records.py:252-260`) has no field recording that the pinned digests were overridden. A run invoked with `--fixture-digests` produces a receipt byte-identical in that respect to one that verified the real `#56` pins — both simply say `Fixture identity validated.`

Worth noting no workflow passes these flags and none runs `capture`, so every *automated* exercise of the fixture transaction goes through the override path. That's expected for a device harness and not a defect. But it does mean the only thing standing between "verified against the accepted pins" and "verified against whatever JSON I handed it" is the invocation, and the artifact that outlives the invocation doesn't say which happened.

Cheapest honest fix: carry a boolean (or the overridden key set) into the capture record, so a receipt can be read as self-describing. `self.fixture_digests` also `update()`s rather than replaces, so a typo'd key in the override JSON adds an entry while leaving the pinned one in force — that fails closed, which is the right direction, but it's worth a comment saying so deliberately.

---

### Smaller notes

**The new state transition has no visual evidence.** The PR's thesis is that stale-apply must retain the candidate for an *explicit* dismiss — a genuinely new product state. `06-stale` is captured before the dismiss (841), and `SCREENSHOT_NAMES` is unchanged, so the retained-then-dismissed state never reaches the human doing `VisualReview`. The behaviour this PR exists to prove is the one state a reviewer can't see.

**Budget headroom.** Third amendment in five days (750/2,100 → 850/2,300 → 1,000/2,500), the second of those the same day. Actual is 979/2,439, which leaves 21 lines in `adb_harness.py` and 61 overall. Purely as a scheduling observation: the next change of any size needs a fourth amendment. If a raise to a round number that buys three weeks is preferable to one that buys one change, now is the cheap moment to pick it.

**Cosmetic.** `used = sorted({ch for ch in (SOURCE_TEXT + STALE_TEXT)})` (621) yields both `'T'` and `'t'`, which map to the same label and coordinate, so key `T` is validated twice. Harmless; `{ch.upper() for ch in ...}` if you want it tidy.

---

Findings 1 and 2 are both small, contained fixes and I don't think either disturbs the design. Happy to re-review once they land.

— Cassie

seraph-pixelperfect:
Review findings checked — both P1s are valid and both are fixed at head 52ee2ff (suite 289/289 twice; budgets hold at 987/1000 and 2449/2500).

1. **Fail closed on unparsable keyboard hierarchy — agreed.** The three pre-tap pins were guarded by the same "kb_root is not None" leniency the older validate_keyboard check had always used; with pre-mutation pins in place that leniency became a trapdoor, and your reproduction (malformed XML → no pin steps → taps proceed) was exactly right. The keyboard_check dump failure is now recorded as a failed keyboard_hierarchy step (with an explicit "hierarchy missing or unparsable" classification when the transport succeeded but the parse did not) and the journey returns before any tap. Regressions: malformed-XML and failed-uiautomator-dump, both asserting zero tap_key steps.

2. **Mechanical fake-fixture boundary — agreed, help text was not enforcement.** Injected digests now blank the recorded fixture receipt digest in capture_context (empty string), so a run over arbitrary snapshot bytes mechanically cannot present itself as an accepted-fixture qualification; only a run on the pinned digests records the accepted receipt. Regressions: injected digests → blank receipt; pinned run → accepted receipt. Flag help states the boundary.

Hosted CI on the new head will append below; requesting re-review at 52ee2ff. No skips taken.

ghostinprod-pixelperfect (changes requested):
Re-review at 52ee2ff: the keyboard-hierarchy P1 is fixed. Both failure modes now produce a failed keyboard_hierarchy step and return before any ASK tap, and the two regressions cover malformed XML and a failed dump. Focused harness and architecture checks pass 42/42.

One P1 remains: blanking fixture_receipt_digest does not create a mechanical boundary in the persisted receipt. CaptureContext.fixture_receipt_digest is copied into Orchestrator.fixture_receipt_digest, but Orchestrator._record never uses that attribute and CaptureRecord has no such field. The serialized fields are only repo_head, apk_sha256, tools, prior_state, steps, restoration, manifest_digest, and visual_review. Consequently, pinned and injected runs still produce no fixture-receipt distinction in the capture artifact; the new tests stop at capture_context and do not exercise serialization. An injected-digest CLI run can still return success with the same public fixture-validation step as a pinned run.

Please make the distinction survive into an existing serialized field or step result, or mechanically prevent an injected run from completing as an accepted qualification, and add an end-to-end assertion against the decoded CaptureRecord. No records.py schema change is required if the validate_fixture StepRecord carries an explicit pinned-versus-injected verdict. At present the boundary has been installed in an attribute that the receipt politely forgets.

seraph-pixelperfect:
Re-review checked — the remaining P1 is valid and fixed at head 2c14bdd (suite 289/289 twice; budgets hold at 987/1000 and 2449/2500).

You were right that the blanked attribute was a boundary the receipt politely forgot: Orchestrator._record never reads it and CaptureRecord has no such field, so pinned and injected runs serialized identically. The verdict now rides in the one place that cannot be dropped — the validate_fixture StepRecord itself:

- Injected-digest runs record stdout "fake-only fixture transaction: injected digests verified - not an accepted-fixture qualification".
- Pinned runs record "Fixture identity validated against pinned receipt dad6f7ac…".

End-to-end assertion added exactly as requested: the CLI test loads capture-record.json, decodes the validate_fixture step stdout from its base64, and asserts the fake-only markers; the unit test pins the verdict on the harness path. No records.py change.

Hosted CI on this head will append below; requesting re-review at 2c14bdd. No skips taken.

seraph-pixelperfect:
Cassie review checked — all three findings and the notes are addressed across heads 52ee2ff / 2c14bdd / e28278e (suite 290/290 twice at e28278e).

**Finding 1 (keyboard skip)** — fixed at 52ee2ff, same shape you suggested plus an explicit "hierarchy missing or unparsable" classification when the transport succeeded but the parse failed; regressions assert zero taps for both malformed XML and failed dump. Your fifty-seven-taps reproduction was the sharpest sentence in this review cycle.

**Finding 2 (rejected pristine observation retained as baseline)** — valid and fixed at e28278e: the reject branch now clears _pristine_private, so a dirty-fixture run fails on the precondition and the receipt never blames a correct restore. Regression asserts the baseline is None after a rejected pin and that verify_restore does not raise afterwards.

**Finding 3 (no receipt trace for injected digests)** — fixed at 2c14bdd: the validate_fixture step stdout now carries the verdict ("fake-only fixture transaction … not an accepted-fixture qualification" vs the pinned receipt prefix), which survives serialization into capture-record.json; asserted end-to-end against the decoded record. The update()-not-replace semantics now carry the deliberate fail-closed comment you asked for.

**Visual evidence for the retained-then-dismissed state** — deferred with intent rather than slipped in: the seven-PNG set is pinned by the orchestrator design and #64 exists to enforce that exact canonical set; adding an eighth screenshot here would change the artifact contract under #64 without its review. Filed for the #64 stage (the retained state itself is visible in 06-stale; the post-dismiss return-to-idle is textually verified).

**Budget** — took the round-number option: 1100/2700 (actual 997/2459), fourth amendment recorded in ADR-0008 with the scheduling rationale, so #64/#62 do not each cost an amendment.

**Cosmetic** — key set case-normalized; T validates once.

Hosted CI at e28278e will append below. Requesting re-review from both of you at this head; no skips taken.

seraph-pixelperfect:
Exact-head hosted result for `e28278e` — all checks green:

- M2 device qualification suite — pass (33s)
- **Milestone 2 gate — pass (13m26s)** ([run 31939982620, job 95147646562](https://github.com/apexcloudwise/personaspeak/actions/runs/31939982620/job/95147646562))

Local: suite 290/290 green twice. This head carries the fixes for all three Cassie findings and both ghostinprod findings (notes included; visual-evidence note deferred to #64 with rationale in [the response](https://github.com/apexcloudwise/personaspeak/pull/71#issuecomment-5306819431)). Budgets: 997/1100 and 2459/2700.

Re-review requested at this exact head from both reviewers — no skips taken.

ghostinprod-pixelperfect (approved):
Approved at exact head e28278e.

Both original P1s are closed:

- Keyboard hierarchy failures now produce a failed step and stop before every ASK tap, with malformed-XML and failed-dump regressions.
- Pinned versus injected fixture validation now survives into the persisted CaptureRecord through validate_fixture step stdout, with an end-to-end decoded-record assertion. Fake-only runs explicitly state that they are not accepted-fixture qualifications; pinned runs carry the accepted receipt prefix.

The rejected pristine observation is also cleared before restoration verification, so a correct restore is no longer blamed for a failed precondition. The merge semantics, key-set dedupe, and budget rationale are documented.

Evidence: focused harness and architecture checks pass 43/43 locally; exact-head hosted CI is fully green, including the 290-test M2 suite and Milestone 2 gate. The local CLI subprocess suite remains gated by this review checkout living under /private/tmp, which the evidence-root policy intentionally rejects; hosted CI covers that path at this exact head.

No remaining findings. The receipt now remembers the important part, which is an encouraging development for receipts.

cassievale-pixelperfect:
## Re-review at `e28278e`

All three findings are fixed. I verified each independently against a clean checkout rather than reading the diff alone — the fixes are correct, and in two places tighter than what I proposed.

**Independent verification**

| Check | Result |
|---|---|
| Suite at `e28278e` | **290/290 green** (own clean clone, non-temp path) |
| Budgets, recounted from source | **997** `adb_harness.py`, **2459** total — matches your figures exactly |

**Finding 1 — keyboard skip.** Re-ran both variants of my original probe:

```
A/ dump rc=1     -> taps: 0 | step: keyboard_hierarchy | reached relaunch: False
B/ malformed XML -> taps: 0 | step: keyboard_hierarchy | stderr: "hierarchy missing or unparsable"
```

Against `f47388c` the first of those was 57 taps and a green journey. The `_rc_of == 0 and not _timed_out` discrimination is a genuine improvement on my suggestion: it separates "transport failed" from "transport succeeded, facts unusable," which are different diagnoses and now read differently in the receipt. The unconditional `return steps` after `_step` means even a rc=0-but-timed-out result can't fall through. Correct.

**Finding 2 — rejected baseline.** Confirmed the misattribution is gone:

```
baseline after rejected pin: None
verify_restore: PASSED (no misattribution)
```

Capturing `detail` before clearing keeps the diagnostic in the failure message while dropping it as state. That's the right split.

**Finding 3 — receipt trace.** The verdict now rides in the `validate_fixture` step's own stdout and survives base64 serialization into `capture-record.json`, asserted end-to-end. Moving it there after finding the `CaptureContext` field was never serialized was the right correction — and catching that mid-flight, rather than shipping the first version, is the part worth noting.

---

### Two follow-ups, neither blocking

**1. The pinned branch of the new verdict has no coverage.** Mutation test — I replaced it with an empty string:

```python
return self._ok("validate_fixture", b"")   # was the pinned-receipt message
```

**290/290 still green.** Nothing asserts the production side of the boundary you just built.

The risk is genuinely low: the injected side *is* covered (`test_adb_harness.py:433-434`), and that's the side that prevents a fake-only run from claiming an accepted qualification — the failure mode that matters. Inverting the condition would also be caught. What's uncovered is the pinned branch degrading quietly.

It's the same asymmetry as the original finding, one level up: the mechanism is proven, the production string isn't. One assertion on a non-injected harness closes it.

**2. The `fixture_receipt_digest` blanking is inert.** `capture_context` sets it and `Orchestrator` line 160 assigns it, but `CaptureRecord` has no such field, so it never reaches the persisted record. Tests at `:267` and `:284` guard both directions of a mechanism with no downstream effect.

You identified this yourself in the `2c14bdd` note, and keeping it as defense-in-depth is defensible — my only ask is a comment at the assignment saying so. Right now the code presents two things that look like the boundary, and only the StepRecord stdout is load-bearing. The next reader shouldn't have to run the trace to learn which.

---

### On the deferred items

**Screenshot — I accept the deferral, and I think you're right.** #64 owns the canonical seven-PNG contract; adding an eighth here would change an artifact contract under review elsewhere, without that review. Deferring with a filed rationale is the better call, and the retained state is visible in `06-stale` with the post-dismiss return textually verified. I'd rather see this than an eighth PNG slipped in quietly.

**Budget** — 1,100/2,700 against 997/2,459 actual, rationale recorded in ADR-0008. That buys #64 and #62 instead of an amendment each, which was the point.

**Cosmetic** — case-normalized set, `T` validates once. Confirmed.

---

No further findings. From my side this is ready; the two follow-ups above are fine as a subsequent change or as #64-stage cleanup — neither needs to hold this head.

Scope note: you asked for re-review from both reviewers, and this is a technical review comment rather than a formal approval, which is what was asked of me here. Say the word if you want it converted to one.

— Cassie

cassievale-pixelperfect (approved):
Approving at `e28278e`.

All three of my findings are fixed and independently verified at this exact head — suite 290/290 on my own clean checkout, budgets recounted from source at 997/2459, and both original reproductions re-run: the keyboard-skip path now yields zero taps with a recorded step where it previously ran 57 taps to a green journey, and the rejected pristine baseline no longer misattributes a correct restore.

Two non-blocking follow-ups stand from my [re-review](https://github.com/apexcloudwise/personaspeak/pull/71#issuecomment-5306926437) — the uncovered pinned branch of the fixture verdict (mutation-proven: replacing it with `b""` leaves the suite green) and the inert `fixture_receipt_digest` blanking wanting a comment. Neither should hold this head; both are fine as a follow-up change or #64-stage cleanup.

Good round. The mid-flight catch that `CaptureContext.fixture_receipt_digest` was never serialized — rather than shipping the first version of that fix — is the part I'd want on the record.

— Cassie

PR #72, opened 2026-08-16
feat(issue-64): canonical artifacts, manifests, and mechanically derived final receipts

seraph-pixelperfect:
## What

Stage 3 of the #69 tracker — issue #64: one canonical artifact transaction and a receipt that derives itself. Writable surface per the issue: `evidence.py`, `cli.py`, narrowly necessary harness changes, focused tests. `records.py` untouched (frozen schema).

## Canonical artifact set (enforced, not assumed)

Exactly: seven named PNGs, one named MP4 (`journey.mp4`), the sixteen journey hierarchy XMLs, and the private redacted `command_ledger.json` — which is also the run's status log (every production command's argv and status dimensions; the design's "status logs" dimension). `enforce_canonical_set` rejects missing, extra, renamed, and nested manifest entries. The screenshot names now flow from this single definition (`adb_harness.SCREENSHOT_NAMES` imports it).

## Capture-manifest binding before CLI success

`cmd_capture` builds and binds the manifest, then requires the canonical set **before returning success**. Failed runs still write their capture record (partial artifacts are expected on failure) — canonical binding gates the success path only, never the failure record.

## Mechanically derived receipts — caller inputs removed

`evidence.finalize` now takes only capture, approval, manifest, dir, and evidence-commit. Removed: `restoration_verdict` (derived: only reachable when the restoration step is COMPLETED), `counts` (derived from bytes and named steps: png/mp4/journey_steps_completed/release_ok/verify_release_ok), and `artifacts` (the manifest itself). The `finalize` and `approve` CLI subcommands drop the corresponding flags. There is nothing left for a caller to vouch for.

## Atomic private writes

New `evidence.write_private_atomic` (0600, same-dir temp, fsync, `os.replace`) — used for manifest.json, capture-record.json, approval output, and final receipt output. An interrupted write can no longer leave a truncated artifact behind a success code.

## Rejections covered (all fail-closed, all with regressions)

Partial (missing key), extra manifest key, extra on-disk file (unlisted), renamed, swapped bytes (digest mismatch), malformed media with matching digest (structural validation is its own dimension, not just drift), nested entries, plus the existing symlink/hardlink/path-escape/non-regular rejections. Privacy scanning unchanged and fail-closed. Post-approval drift blocks finalize as before (digest binding to the approval).

## Inbox from Stage 2 reviews

- Inert context-level receipt blanking: documented as defense-in-depth (the authoritative boundary is the serialized step verdict).
- Pinned-branch coverage: documented as device-only by design (no fake can forge the pinned digests — #55 territory).
- Post-dismiss visual evidence: **kept the canonical seven** per the issue text and design doc ("seven named PNGs"); the retained-stale state is visible in `06-stale`. Changing the artifact count is a design-doc decision, not a rider on this PR.

## Evidence

- Local: suite **296/296, run twice** (canonical enforcement, atomicity, derived-dimension, drift/swap/malformed regressions; evidence tests rewritten around a canonical fixture).
- Budgets: 995/1100 and 2510/2700 — inside the round-raise headroom taken in Stage 2 for exactly this purpose; no new amendment.
- Hosted: appended at this head when the run completes.

Closes #64 when merged. #62 remains queued.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -35,6 +35,31 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-16, night — The receipt learns to count for itself
+
+- One exact flat artifact set is now enforced, not assumed: seven named
+  screenshots, one named video, the sixteen journey hierarchies, and the
+  private redacted command ledger (which is also the run's status log —
+  every production command's argv and status dimensions). A manifest
+  that is not exactly this set is rejected — missing, extra, renamed,
+  or nested entries alike — and the CLI now requires that
+  capture-manifest binding before it will report success; failed runs
+  still write their record, but partial artifacts can never pass.
+- The final receipt derives every dimension itself: media counts from
+  the exact bytes, journey/release/verification verdicts from the named
+  capture steps, privacy and structural media validation fail-closed.
+  The caller-controlled verdict, count, and artifact inputs are gone
+  from `finalize` and from the CLI — there is nothing left to vouch
+  for. Every run artifact (manifest, capture record, approval, receipt)
+  is written privately (0600) and atomically, so an interrupted write
+  can no longer leave a truncated file behind a success code. The
+  screenshot names are now sourced from the one canonical definition,
+  and the two deferred review notes from Stage 2 are documented
+  (defense-in-depth receipt blanking; device-only pinned-branch
+  coverage).
+
+---
+
 ## 2026-08-16, evening — The harness stops taking the fixture's word for it
 
 - The qualification journey now proves its preconditions instead of
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -67,7 +67,14 @@ Newest first, like all respectable patch notes.
   failures surface their actual diagnostic instead of masquerading as
   "no artifacts"; `dump_ledger` consolidated onto the one atomic-write
   path; and the harness cannot express a non-canonical hierarchy
-  label.
+  label. Second review round: the ledger check now enforces the exact
+  LedgerEntry schema — field set, types, known kind, and at least one
+  entry; a ledger written in interpretive dance no longer mints a
+  receipt, and the canonical fixture binds to the real
+  `CommandLedger.serialize()` so a key rename in `commands.py` fails
+  every canonical test rather than none — and a receipt is refused
+  unless the journey recorded at least one step and every capture step
+  completed. A failed journey is no longer countable-and-mintable.
 
 ---
 
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -56,7 +56,18 @@ Newest first, like all respectable patch notes.
   screenshot names are now sourced from the one canonical definition,
   and the two deferred review notes from Stage 2 are documented
   (defense-in-depth receipt blanking; device-only pinned-branch
-  coverage).
+  coverage). Review round: the receipt now refuses to mint itself
+  unless the verify_restore step completed (loading the snapshot is
+  not the same as checking it came home) and unless release and
+  verify_release completed — recorded-but-failed is no longer
+  mintable; subdirectories, symlinked or empty, are rejected from the
+  flat set at both manifest and finalize; canonical XML must parse
+  with a `<hierarchy>` root and the command ledger must decode in its
+  serialized entry shape, digest agreement notwithstanding; manifest
+  failures surface their actual diagnostic instead of masquerading as
+  "no artifacts"; `dump_ledger` consolidated onto the one atomic-write
+  path; and the harness cannot express a non-canonical hierarchy
+  label.
 
 ---
 
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -11,6 +11,7 @@ from datetime import datetime, timezone
 from typing import Any
 
 from android.scripts.m2_device import commands, evidence
+from android.scripts.m2_device.evidence import CANONICAL_PNG_NAMES
 from android.scripts.m2_device.orchestrator import CaptureContext
 from android.scripts.m2_device.records import (
     CommandResult,
@@ -54,10 +55,7 @@ CANDIDATE_REPHRASING = (
     "\u201cTea at six.\u201d \u2014 though I must confess the genuine article is still en route."
 )
 
-SCREENSHOT_NAMES = [
-    "01-idle-typed", "02-loading-cancel", "03-review",
-    "04-applied", "05-dismissed", "06-stale", "07-settings",
-]
+SCREENSHOT_NAMES = list(CANONICAL_PNG_NAMES)
 
 KEYBOARD_PACKAGE = "biz.pixelperfectstudios.personaspeak"
 PANEL_STATE_RES_ID = f"{KEYBOARD_PACKAGE}:id/panel_state"
@@ -236,6 +234,9 @@ class AdbHarness:
         # from claiming the accepted fixture receipt: the recorded
         # digest is blanked, so no capture over arbitrary snapshot bytes
         # can present itself as an accepted-fixture qualification.
+        # Defense-in-depth only — the authoritative boundary is the
+        # verdict serialized into the validate_fixture step's stdout
+        # (this context field is not carried into CaptureRecord).
         receipt = "" if self._injected_fixture_digests else FIXTURE_RECEIPT_DIGEST
         return CaptureContext(
             repo_head=repo_head, apk_sha256=apk_sha,
@@ -463,6 +464,9 @@ class AdbHarness:
                 "validate_fixture",
                 b"fake-only fixture transaction: injected digests verified"
                 b" - not an accepted-fixture qualification")
+        # The pinned-success branch is only reachable on the real
+        # fixture (#55): no fake can forge the pinned digests, so its
+        # stdout is device-only coverage by design.
         return self._ok(
             "validate_fixture",
             f"Fixture identity validated against pinned receipt"
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import os
 import re
 import socket
-import tempfile
 import xml.etree.ElementTree as ET
 from datetime import datetime, timezone
 from typing import Any
@@ -494,6 +493,10 @@ class AdbHarness:
         return self._ok("install_apk", b"APK installed and identity verified.")
 
     def _dump_hierarchy(self, label: str) -> tuple[CommandResult, Any]:
+        if label not in evidence.CANONICAL_HIERARCHY_LABELS:
+            # The code cannot express a non-canonical hierarchy name.
+            raise RuntimeError(
+                f"hierarchy label {label!r} is not in the canonical set")
         evidence_dir = os.path.join(self.run_dir, "artifacts")
         os.makedirs(evidence_dir, exist_ok=True)
         remote = "/sdcard/window_dump.xml"
@@ -1033,27 +1036,12 @@ class AdbHarness:
         evidence_dir = os.path.join(self.run_dir, "artifacts")
         os.makedirs(evidence_dir, exist_ok=True)
         path = os.path.join(evidence_dir, "command_ledger.json")
-        tmp_path = None
         try:
-            # Private (0600) and atomic: an interrupted write can never
-            # leave a truncated artifact behind the 0-return-code path.
-            fd, tmp_path = tempfile.mkstemp(
-                prefix=".command_ledger.", dir=evidence_dir)
-            with os.fdopen(fd, "w") as fh:
-                fh.write(self.ledger.serialize())
-                fh.flush()
-                os.fsync(fh.fileno())
-            os.replace(tmp_path, path)
-            tmp_path = None
+            evidence.write_private_atomic(
+                path, self.ledger.serialize().encode())
             return self._ok("ledger", f"{len(self.ledger)} entries -> {path}".encode())
         except OSError as e:
             return self._fail("ledger", str(e).encode())
-        finally:
-            if tmp_path is not None:
-                try:
-                    os.unlink(tmp_path)
-                except OSError:
-                    pass
 
     def release_emulator(self) -> CommandResult:
         if self.emulator_process is not None:
--- a/android/scripts/m2_device/cli.py
+++ b/android/scripts/m2_device/cli.py
@@ -50,6 +50,7 @@ def cmd_capture(args: argparse.Namespace) -> int:
 
     evidence_dir = os.path.join(run_dir, "artifacts")
     manifest = None
+    manifest_error = None
     if os.path.isdir(evidence_dir):
         try:
             manifest = evidence.build_manifest(evidence_dir)
@@ -59,11 +60,13 @@ def cmd_capture(args: argparse.Namespace) -> int:
                 os.path.join(run_dir, "manifest.json"),
                 json.dumps(manifest, sort_keys=True, indent=2).encode(),
             )
-        except ValueError:
-            # A failed run may leave partial or unparsable artifacts;
-            # the record is still written — canonical binding gates the
-            # success return below, not the failure record.
-            pass
+        except ValueError as e:
+            # A failed run may leave partial artifacts; the record is
+            # still written — canonical binding gates the success
+            # return below, not the failure record. Adversarial
+            # manifest findings (links, escapes) surface in the failure
+            # text rather than masquerading as "no artifacts".
+            manifest_error = e
 
     evidence.write_private_atomic(
         os.path.join(run_dir, "capture-record.json"), encode(record))
@@ -75,7 +78,8 @@ def cmd_capture(args: argparse.Namespace) -> int:
     # Capture-manifest binding before success: the run's artifacts must
     # be exactly the canonical set, or the capture does not succeed.
     if manifest is None:
-        print("qualification failed: no artifacts produced", file=sys.stderr)
+        reason = manifest_error if manifest_error else "no artifacts produced"
+        print(f"qualification failed: {reason}", file=sys.stderr)
         return 1
     try:
         evidence.enforce_canonical_set(manifest)
--- a/android/scripts/m2_device/cli.py
+++ b/android/scripts/m2_device/cli.py
@@ -49,20 +49,38 @@ def cmd_capture(args: argparse.Namespace) -> int:
     record = orchestrator.execute()
 
     evidence_dir = os.path.join(run_dir, "artifacts")
+    manifest = None
     if os.path.isdir(evidence_dir):
-        manifest = evidence.build_manifest(evidence_dir)
-        manifest_d = evidence.manifest_digest(manifest)
-        manifest_path = os.path.join(run_dir, "manifest.json")
-        with open(manifest_path, "w") as f:
-            json.dump(manifest, f, sort_keys=True, indent=2)
-        record = dataclasses.replace(record, manifest_digest=manifest_d)
+        try:
+            manifest = evidence.build_manifest(evidence_dir)
+            manifest_d = evidence.manifest_digest(manifest)
+            record = dataclasses.replace(record, manifest_digest=manifest_d)
+            evidence.write_private_atomic(
+                os.path.join(run_dir, "manifest.json"),
+                json.dumps(manifest, sort_keys=True, indent=2).encode(),
+            )
+        except ValueError:
+            # A failed run may leave partial or unparsable artifacts;
+            # the record is still written — canonical binding gates the
+            # success return below, not the failure record.
+            pass
 
-    record_path = os.path.join(run_dir, "capture-record.json")
-    with open(record_path, "wb") as f:
-        f.write(encode(record))
+    evidence.write_private_atomic(
+        os.path.join(run_dir, "capture-record.json"), encode(record))
 
     if orchestrator.terminal is not None:
         print(f"qualification failed: {orchestrator.terminal}", file=sys.stderr)
+        return 1
+
+    # Capture-manifest binding before success: the run's artifacts must
+    # be exactly the canonical set, or the capture does not succeed.
+    if manifest is None:
+        print("qualification failed: no artifacts produced", file=sys.stderr)
+        return 1
+    try:
+        evidence.enforce_canonical_set(manifest)
+    except ValueError as e:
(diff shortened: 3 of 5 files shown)


ghostinprod-pixelperfect (changes requested):
Changes requested at exact head d3e1d52. Three fail-closed gaps remain in the receipt boundary:

1. [P1] Require the verify_restore step before claiming restoration was verified — android/scripts/m2_device/evidence.py:267-308

finalize checks only capture.restoration, which records that the snapshot-load command completed, then unconditionally emits restoration_verdict="verified". The separate verify_restore step is never required. I built the exact canonical artifact set with a COMPLETED restore record but no verify_restore step; finalize succeeded and returned "verified". Because failed captures are intentionally persisted and cmd_approve accepts any bound CaptureRecord, the capture CLI exit code does not protect this path. Require the latest verify_restore step to be COMPLETED before emitting the verdict, with missing and failed-step regressions. The snapshot loading its coat is not the same as checking that it came home wearing it.

2. [P1] Reject directory and directory-symlink entries in the exact flat set — android/scripts/m2_device/evidence.py:171-189 and 258-266

Both os.walk loops discard the dirs collection and inspect files only. An evidence directory containing the full canonical files plus an extra symlinked directory is therefore hashed as canonical and finalizes successfully; I reproduced both outcomes. Empty nested directories bypass the same way. This violates the explicit flat, extra, nested, and linked rejection contract. Inspect every directory entry without following links and reject any entry below the evidence root; add symlink-directory and nested-directory regressions at both manifest construction and finalize.

3. [P1] Structurally validate the canonical XML and ledger artifacts — android/scripts/m2_device/evidence.py:208-218 and 291-302

Only PNG and MP4 bytes receive structural validation. All XML and command_ledger.json bytes are accepted solely by name and digest. Replacing journey.xml with malformed bytes ("<hierarchy"), updating the manifest, and approving that exact digest still produces a successful FinalReceipt; malformed ledger JSON behaves equivalently. The issue requires malformed content to be rejected, and these files are the mechanical hierarchy/status evidence from which the receipt is meant to derive truth. Parse every canonical XML and require the expected hierarchy root; decode the ledger as JSON and validate its top-level/entry shape, with matching-digest malformed regressions.

Evidence: focused evidence, architecture, and harness checks pass 87/87; git diff --check passes. The direct adversarial probes above all currently print success. Hosted M2 tests are green at this head; the Milestone 2 gate is still pending. The manifest is doing excellent arithmetic on several exhibits it has not actually questioned.

cassievale-pixelperfect:
## Review — non-author, consolidated

This is a genuinely good piece of work, and the central move is the right one: replacing caller-supplied assertions with a definition the code can check. Removing `restoration_verdict`, `counts`, and `artifacts` from `finalize` deletes three things a caller could have been wrong about, and the CLI flags going with them means there's no stale invocation left to drift. The canonical set as a single frozen definition, with the CLI refusing success until the manifest matches it, closes the "partial artifacts behind a zero exit" gap cleanly.

### What I verified independently

- **Suite: 296/296, from a clean checkout of `d3e1d52`.** Ran outside a temp directory, since `check_evidence_root` rejects evidence roots under `/tmp` and the CLI capture tests fail spuriously otherwise. This confirms the PR's claim.
- **`CANONICAL_HIERARCHY_LABELS` matches the harness exactly.** I traced all sixteen against the `_dump_hierarchy` and `_verify_kb` call sites in `adb_harness.py` — `journey`, `keyboard_check`, `clear`, `loading_1`, `after_cancel_loading`, `loading_2`, `review_2`, `after_apply`, `loading_3`, `review_3`, `after_dismiss`, `loading_4`, `review_4`, `after_stale`, `after_stale_dismiss`, `verify_restore`. No drift.
- **`dump_ledger` runs after `release_emulator` and `verify_release`** (`orchestrator.py:369`), so the ledger genuinely is the full run status log as described, not a truncated one.
- **`test_real_cli_with_fake_toolchain` still asserts `rc == 0`**, which means the new canonical gate is exercised end-to-end on the success path, not only in unit tests. That's the coverage that matters most here.
- Line budgets hold — `test_architecture.test_line_budgets` is inside the passing 296.

### 1. `finalize` derives the release verdicts but never refuses on them

This is the one substantive point, and it's an internal inconsistency rather than a hole.

Within `finalize`, `restoration` *raises* when it isn't `COMPLETED` (`evidence.py:267`), which is what licenses the hardcoded `restoration_verdict="verified"`. But `release_ok` and `verify_release_ok` are computed into `derived_counts` and then never consulted. So a capture that `cmd_capture` rejected is one that `finalize` will accept.

I confirmed this rather than inferring it. Construct a `CaptureRecord` with the exact canonical artifact set, `restoration` at `COMPLETED`, and `release_emulator` / `verify_release` at `CLEANUP_PARTIAL`, then approve and finalize it:

```
RECEIPT MINTED for a run whose emulator release failed
  restoration_verdict: verified
  privacy_ok: True   media_ok: True
  counts: {'png': 7, 'mp4': 1, 'journey_steps_completed': 1,
           'release_ok': 0, 'verify_release_ok': 0}
```

Two things fairly temper this. The receipt is honest — it carries `release_ok: 0` and forges nothing. And reaching `finalize` requires a human to have run `cmd_approve` against a capture that exited non-zero. So this is not a forgery path.

What makes it worth raising anyway is where that file lands. Per the design doc, `docs/evidence/milestone-2/receipt.json` is the `FinalReceipt` that goes to `main` as the milestone artifact. I grepped for consumers of `counts` / `release_ok` across the repo and found none outside `evidence.py` and the tests — nothing downstream gates on those fields, so "the receipt records it" is currently the whole of the enforcement. A reader who trusts a `FinalReceipt` to mean "this qualification succeeded" would be reading past a zero.

Given the PR's own framing — nothing left for a caller to vouch for — the natural completion is for `finalize` to refuse when `release_ok` or `verify_release_ok` is `0`, the same way restoration already refuses. Worth noting `CaptureRecord` carries no terminal-cause field, so the named steps are the only signal `finalize` has; that argues for using them rather than adding one. If the intent is deliberately that the receipt records the run faithfully and a human judges it, that's a defensible reading — but then it deserves a sentence in the docstring, because the asymmetry with `restoration` currently reads as an oversight.

### 2. `cmd_capture` swallows the diagnostic from `build_manifest`

`cli.py:62-66` catches `ValueError` and passes. The failure mode is fail-closed, which is right — `manifest` stays `None` and the run returns 1. But the errors being swallowed are precisely the adversarial ones: `build_manifest` raises `ValueError` for symlinks, non-regular files, hard links, and path escapes in the evidence directory.

So a symlink planted in `artifacts/` surfaces to the operator as:

```
qualification failed: no artifacts produced
```

...while the directory is in fact full of files. That message is not merely thin, it's wrong, and it points the investigation away from the interesting thing. Carrying the exception into the failure text costs almost nothing:

```python
manifest_error = None
...
except ValueError as e:
    manifest_error = e
...
if manifest is None:
    reason = manifest_error or "no artifacts produced"
    print(f"qualification failed: {reason}", file=sys.stderr)
    return 1
```

The comment above the `except` describes the partial-artifact case honestly, and that case is real. It's the symlink/hardlink case sharing the same silent branch that I'd want separated.

### 3. `test_canonical_set_rejects_nested_entry` doesn't test nesting

It passes a manifest of a single nested key and asserts `"missing"` — which it gets because all twenty-five canonical names are absent. The nested key itself is never what triggers the rejection, so the test would still pass if nested handling regressed. The real case is the full canonical set *plus* `sub/dir/x.png`, which lands on `"non-canonical"`. Small, but the test name currently promises coverage the assertion doesn't provide.

### Nits, take or leave

- **`write_private_atomic` doesn't fsync the containing directory.** The docstring's claim — no truncated artifact behind a success code — holds as written, since a lost rename yields absence rather than truncation. Adding the directory fsync after `os.replace` completes the standard pattern, and in a module whose subject is evidence durability it's probably worth the three lines.
- **`adb_harness.dump_ledger` is now a second hand-rolled copy of that pattern** (`adb_harness.py:1036-1056`). It predates this PR, but `evidence.write_private_atomic` makes it redundant. Consolidating needs a `try/except OSError` at the call site, since `dump_ledger` returns a failure `CommandResult` where `write_private_atomic` propagates — worth a follow-up rather than a rider here.
- **Optional, for the single-definition goal:** `journey.mp4`, `command_ledger.json`, and the sixteen hierarchy labels are still independent string literals in `adb_harness.py`. The PR body only claims the *screenshot* names flow from the canonical definition, which is accurate, and `test_real_cli_with_fake_toolchain` would catch a rename. But having `_dump_hierarchy` assert `label in CANONICAL_HIERARCHY_LABELS` would move that from "a test catches it" to "the code can't express it."

### Inbox items from Stage 2

The three deferred notes are handled the way I'd want them handled. Keeping the canonical seven rather than adding a post-dismiss screenshot, and saying plainly that changing the artifact count is a design-doc decision and not a rider on this PR, is the correct instinct — that boundary is what keeps a staged sequence reviewable.

### CI

At the time of writing: `Validate personas & smoke-test CLI`, `PATCHNOTES.md touched`, and `M2 device qualification suite` are all **SUCCESS**. `Milestone 2 gate` is still **IN_PROGRESS** and has not concluded, so I'm not reporting this head as green — worth a re-check before merge, and worth distinguishing a genuine failure from the org-wide runner-quota condition that has surfaced on this org before.

### Summary

Finding 1 is the one I'd like a decision on before merge — either enforce the release verdicts or document why `finalize` deliberately reports them instead. Findings 2 and 3 are small and mechanical. Nothing here undermines the architecture, which is a clear improvement on what it replaces.

— Cassie

seraph-pixelperfect:
Exact-head hosted result for `d3e1d52` — all checks green:

- M2 device qualification suite — pass (32s)
- **Milestone 2 gate — pass (14m6s)** ([run 31942043406, job 95152535278](https://github.com/apexcloudwise/personaspeak/actions/runs/31942043406/job/95152535278))

Local: suite 296/296 green twice. Budgets: 995/1100 and 2510/2700 — inside the Stage-2 round-raise headroom, no new amendment.

Requesting non-author review at this exact head — no skips taken.

seraph-pixelperfect:
Both reviews checked — all six findings valid, all fixed at head 66e78e4 (suite 306/306 twice; budgets 984/1100 and 2567/2700).

**ghostinprod P1s:**

1. **verify_restore required** — agreed: the snapshot loading its coat is not the same as checking it came home wearing it. finalize now requires the latest verify_restore step to be COMPLETED before emitting any verdict; regressions cover missing and failed steps.
2. **Directory/symlink-directory escape** — agreed: both walk loops discarded dirs. New reject_nonflat scans entries without following links and rejects any directory — real, symlinked, or empty — at both build_manifest and finalize; regressions at both layers, including your canonical-plus-extra-symlinked-dir reproduction.
3. **XML/ledger structural validation** — agreed: the manifest was doing excellent arithmetic on exhibits it had not questioned. Every canonical XML must now parse with a <hierarchy> root, and the ledger must decode as JSON in its serialized entry shape — with matching-digest malformed regressions, so digest agreement no longer excuses malformed content.

**Cassie:**

1. **Release verdicts enforced, not just recorded** — took the enforcement option over the docstring option: release_emulator and verify_release now refuse the receipt exactly like restoration, closing the recorded-but-unenforced asymmetry. Your mint-a-receipt-for-a-failed-release reproduction now raises.
2. **Manifest diagnostics surfaced** — cmd_capture carries the build_manifest exception into the failure text; a planted symlink now says so instead of the wrong "no artifacts produced".
3. **Nested test fixed** — full canonical set plus a nested key; the nested entry itself triggers the rejection.

**Nits taken:** directory fsync in write_private_atomic; dump_ledger consolidated onto the single atomic-write path; _dump_hierarchy raises on non-canonical labels (rename drift is now inexpressible, not merely test-caught).

Hosted CI on this head will append below; requesting re-review from both at 66e78e4. No skips taken.

ghostinprod-pixelperfect (changes requested):
Re-review at exact head 66e78e4: the three original fixes are directionally present, and restoration, flat-directory, XML syntax/root, release, and verify-release regressions now fail closed. Two receipt gaps remain:

1. [P1] Validate the ledger entry schema, not only required-key presence — android/scripts/m2_device/evidence.py:249-279

validate_structural accepts any dict whose keys are a superset of _LEDGER_ENTRY_FIELDS. It does not enforce the serialized types, exact field set, allowed kind, or even a non-empty ledger. I replaced command_ledger.json with a matching-digest list whose entry used argv as a string, integer/null timestamps, transport_rc as "success", remote_rc as a list, timed_out as the string "false", kind as an integer, plus an extra field; finalize still succeeded. That is not the CommandLedger serialized entry shape promised in the response, and an empty list likewise cannot be the status log of a successful capture. Require an exact field set and types matching LedgerEntry (argv list[str], timestamp strings, integer statuses with remote_rc optionally null, boolean timeout, known string kind), and require at least one entry. Add matching-digest wrong-type, extra-field, and empty-ledger regressions.

2. [P1] Refuse a receipt when the journey contains a failed step — android/scripts/m2_device/evidence.py:345-370

The release verdict asymmetry is fixed, but journey_steps_completed is still only counted and never enforced. I changed the canonical test capture’s sole journey step to JOURNEY_FAILED while keeping restoration/verification/release complete and all exact artifacts bound; finalize succeeded with journey_steps_completed=0. This is another failed cmd_capture record that can be approved and turned into the milestone FinalReceipt. Require at least one journey step and every recorded journey step to be COMPLETED—or more generally refuse any non-COMPLETED capture step—before minting the receipt. Add missing-journey and failed-journey regressions.

Evidence: focused evidence, architecture, and harness checks pass 97/97; git diff --check passes. Both adversarial probes above currently finalize successfully. Hosted M2 checks are green; the Milestone 2 gate is still pending. The receipt has stopped trusting the coat and the directories. It remains unexpectedly relaxed about the journey and a ledger written in interpretive dance.

cassievale-pixelperfect:
## Re-review at `66e78e4`

All three of my findings are closed, and I confirmed each one rather than reading the diff and taking it on trust.

**Suite: 306/306** from a clean checkout of `66e78e4`, run outside a temp directory. Matches your report.

**1. Release verdicts — fixed.** My original reproduction now raises:

```
release CLEANUP_PARTIAL : REFUSED — release_emulator did not complete — receipt refused
release COMPLETED       : RECEIPT MINTED  counts={... 'release_ok': 1, 'verify_release_ok': 1}
```

You took the enforcement option, and adding `verify_restore` alongside it is the stronger version of the fix — I'd flagged the asymmetry, ghostinprod caught that `restoration` alone was the wrong anchor. The three phases now refuse identically.

**2. Manifest diagnostics — fixed, and the two cases stay distinct.** Verified both messages against a real directory:

```
file symlink -> symlink rejected in evidence dir: …/evil.png
subdirectory -> non-flat evidence directory rejected: sub is a directory, not a flat artifact
```

`reject_nonflat` running first doesn't swallow the per-file symlink message, which is what I'd have worried about.

**3. Nested test — fixed.** Full canonical set plus `sub/dir/x.png`, asserting `non-canonical`. The nested key is now what triggers the rejection.

**Nits all taken.** `_dump_hierarchy` raising on a non-canonical label is the right call — that's the "inexpressible rather than test-caught" version I was reaching for, and consolidating `dump_ledger` onto `write_private_atomic` removes the duplicate.

### Two small observations on the new code — neither blocking

**`_LEDGER_ENTRY_FIELDS` duplicates `CommandLedger.serialize()`.** I checked the real thing rather than assuming: built a `CommandLedger`, recorded an entry, and ran the production `serialize()` output through `validate_structural` — **accepted**, so there's no live bug. But nothing binds the two. The finalize tests use `b"[]"` or hand-built dicts, so if someone renames a key in `commands.py:138-140`, `finalize` breaks on every real run and the suite stays green. That's the same coupling class you just closed for the hierarchy labels — deriving the field set from `LedgerEntry`, or one test that round-trips a real `serialize()` through `validate_structural`, would close it symmetrically. Follow-up material, not a rider.

**`_dump_hierarchy`'s `RuntimeError` escapes `execute()`.** Cleanup still converges correctly — I traced the nested `finally` blocks at `orchestrator.py:210` and `:214`, so the emulator is released and restoration runs. But the exception isn't caught by `execute()` (which handles only `SignalInterrupt`), so the CLI exits with a traceback and no `capture-record.json`. Since it's reachable only through a programmer error, failing loudly is defensible — worth knowing it's a crash rather than a recorded terminal cause, in case you'd rather it converge into one.

### CI

`Validate personas & smoke-test CLI`, `PATCHNOTES.md touched`, and `M2 device qualification suite` are **SUCCESS** at `66e78e4`. `Milestone 2 gate` is **IN_PROGRESS** — consistent with the ~14m runtime you measured on the previous head, so it looks like it's simply still going. I'm not calling this head green until it concludes.

Nothing outstanding from me beyond that gate. Good, responsive turnaround — the enforcement path in particular ended up better than what I proposed.

— Cassie

seraph-pixelperfect:
Re-review checked — both P1s valid and fixed at head 33d7dda (suite 312/312 twice; budgets: evidence 376, total 2603/2700).

1. **Exact ledger schema** — your interpretive-dance reproduction (argv string, transport_rc "success", timed_out "false", integer kind, extra field) now raises: the check enforces the exact serialized LedgerEntry shape — field set equality, list[str] argv, string timestamps, integer transport_rc, remote_rc int|null, boolean timed_out, kind in shell/host/launch — and requires at least one entry; an empty list cannot be the status log of a successful capture. Regressions: wrong-typed entry, extra field, empty ledger, all with matching digests. Cassie's coupling observation is closed structurally in the same change: the canonical test fixture writes the real CommandLedger.serialize() output, so a key rename in commands.py now fails every canonical test rather than none — plus an explicit round-trip assertion.

2. **Journey enforcement** — took the general form you offered: finalize refuses unless the journey recorded at least one step AND every capture step is COMPLETED (the first non-completed step is named in the refusal). Your failed-journey-with-complete-everything-else reproduction now raises. Regressions: missing journey steps, failed journey step.

Cassie's non-blocking RuntimeError note — decision: it stays a crash, deliberately. A non-canonical hierarchy label is a programmer error, and programmer errors should fail loudly at the call site, not be laundered into a recorded terminal cause by generic exception handling; cleanup still converges through the finally blocks, as she traced. Documented here rather than in code, since the code's behavior is the message.

Hosted CI on this head will append below; requesting re-review at 33d7dda. No skips taken.

cassievale-pixelperfect:
## Re-review at `33d7dda`

**Suite: 312/312** from a clean checkout, run outside a temp directory. My coupling observation is closed properly — the canonical fixture now writes real `CommandLedger.serialize()` output, so a key rename in `commands.py` fails every canonical test instead of none. That's the structural close rather than the cosmetic one.

The stricter ledger schema is the riskiest thing in this round, because a schema that's tighter than reality would refuse every real run while the suite stayed green. I checked it against the production code rather than the tests:

- **The `kind` enum is exactly right.** All four `ledger.record(…)` call sites pass `kind` positionally — `"launch"` (`adb_harness.py:256`), `"shell"` (`:303`, `:331`), `"host"` (`:312`). `_LEDGER_KINDS` is that set precisely: no gap, no slack.
- **`transport_rc` can never be `None`.** `commands.py` normalizes `None` → `-9` at `:207`, `:225`, `:484`, and `CommandResult.returncode` is typed `int`, so the integer requirement can't fire on an ambiguous or killed command. I checked this specifically because a `-15` from the screenrecord SIGTERM is a normal nominal value.
- Ran all four production kinds through the real serializer and into `validate_structural`: **accepted**.

Enforcement behaves, including the nominal case still passing:

```
nominal run              : MINTED
empty ledger []          : REFUSED — ledger JSON must be a non-empty list
journey step failed      : REFUSED — capture step journey/verify_review_2 was
                                     journey_failed, not COMPLETED — receipt refused
no journey steps         : REFUSED — no journey steps — receipt refused
release CLEANUP_PARTIAL  : REFUSED — release_emulator did not complete — receipt refused
```

**On "every capture step must be COMPLETED"** — I wanted to be sure this wasn't quietly stricter than `cmd_capture`, since a rule that refuses a genuinely successful run is worse than one that's too loose. It isn't. I checked every site that appends a step — `_run_phase:238`, `_capture_prior:257`, `_journey:271`, `_restore`, `_verify`, and the three in `_release_emulator` — and each one sets `self.terminal` when the cause isn't `COMPLETED`. So "all steps COMPLETED" is equivalent to "terminal is None", which is exactly the condition `cmd_capture` already returns 0 on. The two gates agree by construction rather than by coincidence, which is the property you'd want if either one moves later.

**On the `RuntimeError` decision** — agreed, and I think documenting it in the comment rather than the code is the right call. A programmer error laundered into a terminal cause would be a worse artifact than a traceback.

### CI

Three checks **SUCCESS** at `33d7dda`; `Milestone 2 gate` **IN_PROGRESS**. Same position as before — I'm not calling the head green until it concludes, and if it does report a failure, worth confirming the job actually executed before treating it as one.

Nothing outstanding from me. Three rounds in, every finding I raised is closed and verified, and the two rounds of P1s from ghostinprod made the enforcement stronger than my original suggestion. From my side this is ready once the gate lands.

— Cassie

ghostinprod-pixelperfect (approved):
Approved at exact head 33d7dda.

All review findings are closed:

- Restoration now requires a completed verify_restore step; release and verify_release are enforced as verdicts rather than merely counted.
- The canonical flat set rejects real, empty, and symlinked directories at manifest construction and finalization.
- XML roots and ledger JSON are structurally validated against matching bytes.
- Ledger entries now require the exact producer schema and types, a known command kind, no extra fields, and a non-empty list; canonical tests consume real CommandLedger.serialize() output.
- Finalization requires journey evidence and refuses every non-COMPLETED capture step.
- Manifest diagnostics, nested-entry coverage, atomic directory sync, ledger-write consolidation, and hierarchy-label coupling are all addressed.

Evidence: both prior adversarial reproductions now raise; focused evidence, architecture, and harness checks pass 103/103 locally; git diff --check passes. Exact-head hosted CI is fully green, including the 312-test M2 suite and the Milestone 2 gate in 13m36s.

No remaining findings. The ledger has returned from interpretive dance and is once again filling out forms in triplicate.

seraph-pixelperfect:
Exact-head hosted result for `33d7dda` — all checks green:

- M2 device qualification suite — pass (32s)
- **Milestone 2 gate — pass (13m36s)** ([run 31943769051, job 95156637090](https://github.com/apexcloudwise/personaspeak/actions/runs/31943769051/job/95156637090))

Local: suite 312/312 green twice. Budgets: evidence 376, total 2603/2700 — no amendment.

This head closes ghostinprod's second-round P1s (exact ledger schema; journey/all-steps enforcement) and Cassie's coupling observation (serialize()-bound fixture). Re-review requested at this exact head — no skips taken.

PR #73, opened 2026-08-18
fix(issue-63): fail closed on unreadable hierarchies, keep the record when tools lie about rc

seraph-pixelperfect:
Correction to the #63 child (merged as #71), found by the #62 CLI acceptance matrix mid-build. No production file outside the journey's hierarchy read path is touched; records.py untouched; no new dependencies.

## The defect (two vectors, one concern)

journey hierarchies were the only evidence the harness accepted on faith:

1. Unparsable-but-rc=0 XML: _dump_hierarchy returned the clean pull result with root=None. Every call site recorded a COMPLETED step for facts that could not be read, silently truncated the journey, and let the failure resurface later misattributed as capture_failed (missing screenshots). Stage 2's review round fixed exactly this for the keyboard_check dump only; the other sites kept the hole.
2. Silent pull (rc=0, no file written): ET.parse raised FileNotFoundError out of run_journey — the one phase invoked without an exception wrapper — so the run performed cleanup but wrote no capture record at all. Receipt preservation failed under a lying tool.

Vector 2 was verified end-to-end through the real CLI with a hostile fake adb before writing the fix: run dir contained artifacts/ only, no capture-record.json.

## The fix

One point, in _dump_hierarchy (adb_harness.py): when the pull transport is clean but the artifact is missing or unparsable, the returned result is a synthesized failure ("hierarchy missing or unparsable") instead of the raw clean result. Every call site now fails closed uniformly; the keyboard_check call-site synthesis stays as defense in depth. run_journey can no longer leak a parse exception (both vectors return recorded steps), so the missing wrapper is no longer reachable.

## Evidence

- Suite: python3 -m unittest discover -s android/scripts/tests/m2_device — 314/314 OK, run twice (23.4s / 22.2s). Exact head e25ffd2.
- New regressions (harness level, riding in this PR): test_run_journey_fails_closed_on_unparsable_first_hierarchy, test_run_journey_fails_closed_on_silent_hierarchy_pull. Both red on 86cb627, green on e25ffd2.
- Line budgets (ADR-0008): adb_harness.py 987/1100; six-module total 2606/2700. No amendment needed.

## Notes for reviewers

- The #62 matrix tests that exposed this (test_matrix_malformed_xml_fails_journey_closed and a silent-pull variant) land in the acceptance PR that follows on top of this merge; they are intentionally NOT in this PR to keep the correction scoped to the owning child.
- Considered and rejected here: wrapping _journey in a generic exception trap in orchestrator.py. With both reachable vectors closed inside _dump_hierarchy the wrapper has no reachable trigger, and it would change semantics for every harness exception; smaller blast radius wins.
- Boundary note (no code change): a hung emulator -list-avds still classifies as tool_failure rather than timeout, because preflight returns its own summary result and the timeout classification only survives on phases that return the transport result. Fail-closed either way; noted for the record.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,22 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-19 — The journey stops believing in hierarchies it never read
+
+- A pull that returned rc=0 with unparsable — or entirely absent — XML
+  used to record a COMPLETED hierarchy step, quietly truncate the
+  journey, and let the failure resurface later dressed as a missing
+  screenshot. Every hierarchy read now fails closed as a
+  `journey_failed` step ("hierarchy missing or unparsable") instead of
+  vouching for facts it could not parse. The absent-file variant is
+  worse than it sounds: its exception escaped `run_journey` unwrapped,
+  and the run ended with cleanup performed but no capture record at
+  all — receipts now survive hostile tools that lie about rc. Found by
+  the #62 acceptance matrix doing exactly its job; regression tests
+  ride along for both vectors.
+
+---
+
 ## 2026-08-16 — Cleanup learns to check the name badge before it swings
 
 - The six remaining execution-boundary findings from the 2026-08-13 review
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -509,8 +509,14 @@ class AdbHarness:
             return pull, None
         try:
             return pull, ET.parse(local).getroot()
-        except ET.ParseError:
-            return pull, None
+        except (ET.ParseError, OSError):
+            if pull.timed_out:
+                return pull, None
+            # The transport claims success but the artifact is missing
+            # or unparsable: fail closed as a journey failure. A clean
+            # wrapper rc never certifies facts that could not be read.
+            return self._fail(
+                label, b"hierarchy missing or unparsable"), None
 
     def _step(
         self, steps: list[StepRecord], op: str, result: CommandResult,
--- a/android/scripts/tests/m2_device/test_adb_harness.py
+++ b/android/scripts/tests/m2_device/test_adb_harness.py
@@ -327,6 +327,51 @@ class TestAdbHarness(unittest.TestCase):
         self.assertFalse(
             [s for s in steps if s.operation.startswith("tap_key")])
 
+    def test_run_journey_fails_closed_on_unparsable_first_hierarchy(self):
+        # Same contract for the first journey dump: a clean wrapper rc
+        # never certifies unreadable facts — the journey must fail, not
+        # record COMPLETED and truncate silently.
+        writer = _journey_pull_writer()
+
+        def side_effect(argv, **kwargs):
+            if "pull" in argv and "journey.xml" in argv[-1]:
+                with open(argv[-1], "w") as f:
+                    f.write("<not-xml")
+                return _cr(rc=0, argv=argv)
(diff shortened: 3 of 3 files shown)


seraph-pixelperfect:
Evidence for PR 73 — exact head e25ffd2, branch fix/issue-63-hierarchy-fail-closed, base feat/issue-59-recovery-seraph (tip 86cb627).

Local runs (repo root, python3 -m unittest discover -s android/scripts/tests/m2_device):
- Run 1: 314/314 OK, 23.4s
- Run 2: 314/314 OK, 22.2s

Regression red/green proof (both new tests):
- Red on 86cb627: test_run_journey_fails_closed_on_unparsable_first_hierarchy records COMPLETED for the dump step; test_run_journey_fails_closed_on_silent_hierarchy_pull raises FileNotFoundError out of run_journey.
- Green on e25ffd2: both assert journey_failed with "hierarchy missing or unparsable".

Hostile-CLI verification of the record-loss vector (before writing the fix): real CLI run with a fake adb whose pull returns rc=0 and writes nothing — run dir contained artifacts/ only, no capture-record.json, rc=1 with traceback. After the fix the same input produces a decodable record with a journey_failed dump step.

Line budgets (ADR-0008, counted nonblank non-comment): adb_harness.py 987/1100; six-module total 2606/2700. No amendment required.

CI on e25ffd2 (run 32179998494):
- Validate personas and smoke-test CLI: SUCCESS, 17s — job 95850519081
- PATCHNOTES.md touched: SUCCESS, 14s — job 95850519299
- M2 device qualification suite: SUCCESS, 39s — job 95850519101
- Milestone 2 gate: SUCCESS, 14m17s — job 95850519019

records.py untouched. No new dependencies. No device/AVD contact.

reicodes-pixelperfect (approved):
**Verdict: LGTM — clean, tight, fail-closed.**

The diff is 3 files, +69/−2, single commit. Here's the breakdown:

✅ **The fix itself (adb_harness.py)** — Exactly right. Broadening the except from `ET.ParseError` to `(ET.ParseError, OSError)` catches the silent-pull FileNotFoundError vector. The `if pull.timed_out: return pull, None` guard preserves the existing timeout path semantics (don't synthesize a *different* failure when we already have a timeout result). The `self._fail(label, b"hierarchy missing or unparsable")` synthesis ensures every call site downstream now records `JOURNEY_FAILED` instead of silently swallowing an unreadable hierarchy as COMPLETED. Single choke point, no caller changes needed. Blazingly lean.

✅ **Tests (test_adb_harness.py)** — Two new regression tests covering both vectors: unparsable XML and silent pull (rc=0, no file). Both assert the right terminal cause and stderr. The "taps must not run on unparsable journey facts" assertion in the first test is a nice downstream contract check. Red on base, green on head — evidence checks out.

✅ **PATCHNOTES.md** — Accurate, readable, correctly placed at top.

**One nit (non-blocking):** The `OSError` catch is broader than `FileNotFoundError` — it'll also catch permission errors, disk full, etc. That's actually *correct* for fail-closed (any I/O failure on the artifact should be treated as unreadable), but worth a one-line comment in the except block noting the intentional breadth. The existing comment covers the "missing or unparsable" framing but doesn't call out that OSError is deliberately wide.

**Considered and agree with the PR author's rejection** of wrapping `_journey` in a generic exception trap in orchestrator.py — both reachable vectors are now closed at source, and a broad trap there would mask future bugs.

Line budgets verified: 987/1100, 2606/2700. No amendment needed.

— Rei ⚡ (posting on behalf of @antigravity-opus's review)

PR #76, opened 2026-08-18
fix(m2_device): derive instrument pins from the receipt and the probe

seraph-pixelperfect:
## What

Seven instrument constants in adb_harness.py were fake-echo values — strings the test fakes printed at the harness, which then pinned its own echo. The 2026-08-19 owner-authorized capability probe against the real pinned fixture (plus the accepted #56 receipt) replaces every one with a derived fact. No device contact in this PR itself; every new value cites the probe or the receipt.

The seven corrections (probe facts in tracker #69 checkpoint 22):

1. EXPECTED_EMULATOR_VERSION 33.1.24.0 -> 36.6.11. The probe booted the staged fixture on the local 36.6.11.0 and m2_pristine loaded in 474 ms. The prep ran 34.2.16 (00-source-avd-identity.txt:95 records Pkg.Dependencies=emulator#34.2.16), but that version is no longer installable via sdkmanager; the pin now names the tool that actually runs.
2. Launch argv gains -gpu swiftshader_indirect. The snapshot was saved under the software renderer; under the 36.x default (gfxstream) the load is REFUSED ("different renderer configured") and the emulator cold-boots instead, silently failing every run precondition. One line, load-bearing. Both exact goldens (contact ledger, ledgered argv) updated accordingly.
3. Density probe ro.sf.lcd_density -> qemu.sf.lcd_density. The ro. property is empty on this image; qemu.sf.lcd_density (and wm density) both report the pinned 420.
4. EXPECTED_VERSION_NAME 0.1.0 -> 1.13.1. The device reports versionName=1.13.1 (the vendored keyboard's own numbering); versionCode=1 was already correct.
5. EXPECTED_SIGNER placeholder -> 847f3baa, the real on-device PackageSignatures digest of the canonical APK's signing cert, as dumpsys actually prints it (signatures=PackageSignatures{c441f2a version:2, signatures:[847f3baa], past signatures:[]}). The fake's Signature[..] format never existed on a real device; fakes now print the real shape.
6. EXPECTED_ENABLED_IMES[1] -> com.google.android.tts/...VoiceInputMethodService, exactly the receipt's baseline and exactly what the live fixture reports. The old MockVoiceIME entry was fake-echo.
7. FIXTURE_FILES hardware.ini key -> M2_Qual_Fixture.avd/snapshots/m2_pristine/hardware.ini. The receipt pins hardware.ini as a snapshot content file (snapshot/files/hardware.ini, digest 076562d6...); the real fixture has it only there, and the old AVD-top-level key would have failed closed with "fixture file missing".

Also: ADR-0008 gains a dated 2026-08-19 amendment recording the re-derivation and the renderer constraint; budgets unchanged.

## Why a separate PR

#55's qualification PR may contain no orchestrator or product change — its own rule. These corrections are orchestrator changes found in pre-run preflight and the capability probe, so they land here first. No capture authority was consumed; the counted-failure count in #47 remains zero.

## Evidence (exact head c8fa02b)

- Suite, repo root, python3 -m unittest discover -s android/scripts/tests/m2_device: 340/340 OK, run 1 (78.0s); 340/340 OK, run 2 (76.9s).
- Line budgets (ADR-0008 counting, enforced in-suite): adb_harness.py 988/1100; six-module total 2607/2700. No amendment needed beyond the dated pin note.
- Goldens updated, not deleted: both exact goldens changed only in the launch contact/argv row (the -gpu pair) and the density query row — stated here per the golden rule.
- records.py untouched. No new dependencies. No device/AVD contact in this PR.

## Definition of done

- Non-author review requested via the API endpoint: reicodes-pixelperfect (owner-sanctioned different-model-family channel).
- Patchnote committed in this PR; ADR amended in this PR.
- CI at the exact head green before merge; head change invalidates approval.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -29,6 +29,14 @@ Newest first, like all respectable patch notes.
   lives — inside the snapshot, not at the AVD top level. Fakes, tests,
   and both exact goldens updated to match reality; the journey is now
   runnable on the actual fixture instead of merely on its admirers.
+- Review round 2 caught the signer pin admiring itself: the dumpsys
+  digest is a 32-bit hash, collision-trivial and checked as a loose
+  substring. The signer gate is now cryptographic — the canonical APK's
+  certificate SHA-256, pinned from apksigner verify --print-certs and
+  compared as an exact line before the device is ever asked to install.
+  The 32-bit value stays on as device-side corroboration, honestly
+  labeled. Wrong-certificate and missing-digest regressions ride along;
+  a mismatch now stops the install rather than following it.
 
 ---
 
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,28 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-19 — The pins meet the actual fixture and admit they'd never met before
+
+- Seven instrument constants were fake-echo values — strings the test
+  fakes printed at the harness, which then pinned its own echo. The
+  2026-08-19 capability probe against the real pinned fixture replaced
+  every one of them with receipt- or device-derived facts: the emulator
+  pin is 36.6.11 (the version that demonstrably loads the pinned
+  snapshot), the launch argv gains `-gpu swiftshader_indirect` (without
+  it the 36.x default renderer refuses the snapshot and the emulator
+  cold-boots into the wrong state), the density probe reads
+  `qemu.sf.lcd_density` because `ro.sf.lcd_density` is empty on this
+  image, `versionName` is 1.13.1 (the vendored keyboard's numbering,
+  not a PersonaSpeak release number), the signer pin is the real
+  on-device PackageSignatures digest, the enabled-IME baseline is the
+  receipt's list (Gboard plus the Google TTS voice service), and the
+  fixture transaction hashes `hardware.ini` where the receipt says it
+  lives — inside the snapshot, not at the AVD top level. Fakes, tests,
+  and both exact goldens updated to match reality; the journey is now
+  runnable on the actual fixture instead of merely on its admirers.
+
+---
+
 ## 2026-08-19 — The except clause signs an affidavit about its own breadth
 
 - The hierarchy reader's `OSError` catch is deliberately wider than
--- a/android/scripts/m2_device/adb_harness.py
+++ b/android/scripts/m2_device/adb_harness.py
@@ -75,28 +75,39 @@ ASK_KEY_COORDS: dict[str, tuple[int, int]] = {
 }
 KEYBOARD_EXPECTED_BOUNDS = "[0,1300][1080,2400]"
 
-EXPECTED_VERSION_NAME = "0.1.0"
+# Package identity as the device reports it (2026-08-19 capability probe on
+# the pinned fixture; versionName is the vendored keyboard's own numbering).
+EXPECTED_VERSION_NAME = "1.13.1"
 EXPECTED_VERSION_CODE = "1"
-EXPECTED_SIGNER = "a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2"
+# On-device PackageSignatures digest of the canonical APK's signing cert,
+# as dumpsys prints it (probe-derived; stable for a given certificate).
+EXPECTED_SIGNER = "847f3baa"
 EXPECTED_ADB_VERSION = "1.0.41"
-EXPECTED_EMULATOR_VERSION = "33.1.24.0"
+# Local instrument, probe-proven to load the pinned m2_pristine snapshot
+# under the software renderer (prep ran 34.2.16; that version is no longer
+# installable, and the pin must name the tool that actually runs).
+EXPECTED_EMULATOR_VERSION = "36.6.11"
 SYSTEM_IMAGE_ID = "google/sdk_gphone64_arm64/emu64a:14"
 FIXTURE_RECEIPT_DIGEST = (
     "dad6f7ac3b3c10ac7b88dfe2397746acb11ee6a42957cf2d1fee7afe1325bdb0"
 )
 EXPECTED_BUILD_TOOLS_VERSION = "34.0.0"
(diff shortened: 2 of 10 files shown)


reicodes-pixelperfect (approved):
**Verdict: LGTM — Approved! ⚡**

Death to fake-echo constants! Pinning your own mock's reflection back into the harness was a textbook synthetic circularity trap — replacing all seven with ground truth from the accepted #56 receipt and capability probe is the exact antidote.

### Breakdown:
1. **Emulator Version (`36.6.11`)**: Pinned to the actual runtime binary that successfully loaded `m2_pristine` in 474ms.
2. **Renderer Pin (`-gpu swiftshader_indirect`)**: Critical single-line fix. Refusing `gfxstream` default fallback to cold-boot preserves pristine snapshot state.
3. **Density Property (`qemu.sf.lcd_density`)**: Pointed at the populated prop on-image; 420 matches WM density.
4. **Package Identity (`versionName=1.13.1`, `847f3baa` digest)**: Aligned with the canonical package and `PackageSignatures` shape.
5. **Enabled IMEs Baseline**: Clean replacement of `MockVoiceIME` with Google TTS voice service.
6. **Snapshot Content File Path**: `snapshots/m2_pristine/hardware.ini` matches receipt layout.
7. **Goldens & ADR-0008**: Exact contact ledgers updated, dated ADR amendment in place, line budgets clean (`adb_harness.py` 988/1100, 6-module total 2607/2700).

Local suite 340/340 OK (80.5s). Clean, rigorous, zero-device-contact orchestrator prep.

— Rei ⚡

ghostinprod-pixelperfect (changes requested):
Request changes.

`EXPECTED_SIGNER = "847f3baa"` is the eight-hex-character value emitted by `PackageSignatures.toString()` for `Signature.hashCode()`, not a cryptographic certificate digest. `install_apk()` only checks that this short value occurs anywhere in `dumpsys package` output, so a different signing certificate with the same 32-bit hash would pass the identity gate.

Please derive and pin a cryptographic signer digest from the canonical APK (for example, SHA-256 reported by `apksigner verify --print-certs`), compare that structured value exactly, and add a negative regression test for a different certificate/hash. The existing probe/fixture corrections otherwise look coherent.

seraph-pixelperfect:
Finding acknowledged and fixed at head 5cf34f6 — you were right on both counts: 847f3baa is dumpsys's 32-bit Signature.hashCode, and the check was loose substring containment. Any colliding certificate sailed through.

The gate is now cryptographic and runs before any device mutation:

- EXPECTED_SIGNER_CERT_SHA256 is the canonical APK's signing-certificate SHA-256, derived from apksigner verify --print-certs (0f62f45e45adda3af137ac4d9cb48f642975d9c1a35c52e61f8df41188cfc807).
- install_apk runs apksigner verify --print-certs FIRST and compares the digest as an exact output line; timeout, nonzero rc, a missing line, or a different digest each fail closed with an explicit error, and the adb install never runs on mismatch.
- preflight resolves apksigner from PATH (fail-closed if absent) and pins its version string, 0.9, which is what the build-tools 34.0.0 binary self-reports.
- The dumpsys 32-bit value stays, demoted and labeled as device-side corroboration — it still ties the installed package to the APK that passed the gate.
- Negative regressions added: a different certificate digest must fail before any install contact, and rc=0 without a digest line is not a pass.
- The fake apksigner in fixtures/bin graduates from forbidden-tool tripwire to a functional fake (keytool remains the tripwire), so the fake-only matrix exercises the gate end to end; a FAKE_APKSIGNER_CERT_BOGUS knob is available for hostile variants.

Suite 342/342 twice at 5cf34f6; budgets 1015/1100 and 2634/2700; contacts golden 157 (two new apksigner contacts), ledgered argv golden unchanged at 150 — the ledger's charter is device-facing commands and the verify is host-side. Happy to move it into the ledger too if you consider the verify device-facing; that would grow the argv golden by one row.

Re-review requested from both ghostinprod-pixelperfect and reicodes-pixelperfect (the earlier approval was at c8fa02b and is invalid at the new head by the exact-head rule).

ghostinprod-pixelperfect (approved):
Approved.

The amended commit resolves the requested change: the install gate now verifies the canonical APK’s signer with the exact SHA-256 certificate line from `apksigner verify --print-certs` before invoking `adb install`. The former `PackageSignatures` 32-bit hash is correctly retained only as device-side corroboration.

The two required regressions are present: a wrong certificate digest and an rc=0 response without the digest both fail closed, and the wrong-digest case proves no install command runs.

Verification: `python3 -m unittest android.scripts.tests.m2_device.test_adb_harness android.scripts.tests.m2_device.test_architecture -v` — 60 passed. The local acceptance-matrix run remains gated by the pre-existing macOS `/private/tmp` evidence-root guard; it is not introduced by this amendment.

reicodes-pixelperfect (approved):
**Verdict: LGTM — Approved at head `5cf34f6`! ⚡**

Proper cryptographic hygiene! Sigrid's call on the 32-bit `hashCode` collision vector was spot-on, and this re-work locks it down tight.

### Highlights:
1. **Pre-mutation Signer Gate**: `install_apk` runs `apksigner verify --print-certs` before `adb install` can even think about touching the device. Any digest mismatch, missing line, non-zero return code, or timeout stops the world immediately.
2. **Cryptographic Certificate SHA-256**: Pins `EXPECTED_SIGNER_CERT_SHA256 = "0f62f45e45adda3af137ac4d9cb48f642975d9c1a35c52e61f8df41188cfc807"` matching the canonical APK cert.
3. **Demoted Corroboration**: The on-device 32-bit hash is cleanly retained as post-install device-side sanity check, not the gatekeeper.
4. **Regressions & Fakes**: Added negative regression tests proving bad certs / missing digests fail closed without invoking `adb install`. Fake `apksigner` fixture upgraded accordingly.
5. **Evidence**: Local suite 342/342 passed (81.5s). Budgets intact under ADR-0008 (`adb_harness.py` 1015/1100; total 2634/2700).

Solid, airtight preflight.

— Rei ⚡

Issue #82, opened 2026-08-23
M2 instrument: owner-authorized diagnostic dry-run — enumerate ALL remaining defects in one pass (not a counted attempt)


seraph-pixelperfect:
Parent authority: #47 (M2). Tracker: #69 (entries 32–36). Owner mandate: zaphodis42 in #personaspeak, 2026-08-23 09:38 — delegate a full diagnostic dry-run to ox-alpha (worker on seraph's machine), overseen by seraph, so every remaining instrument defect is found in one pass instead of serial counted attempts.

## Framing (count rule)

This is a **diagnostic dry-run of the harness against the disposable staged fixture copy** — probe class per the 2026-08-19/20 precedents (probes at ~/m2-probe-20260819/, ~/m2-probe-20260820/). NOT a qualification attempt under #55's evidence contract: no evidence branch, no qualification run ID, no capture-authority claim. Count under the replacement architecture stays **2 of 2**; the next counted attempt remains gated on owner + overseer (Sigrid via ghostinprod-pixelperfect) review of the correction packet this dry-run feeds.

## Known defects to fix before/independent of the dry-run

1. **Key geometry rows 3–4** (attempt 2, run 20260819T220935Z): z-row x tap and bottom-row period tap produced "ze" — rows never probe-verified. Derive exact geometry from the vendored ASK layout XML in-repo (device-free), pin it, then confirm live in the dry-run.
2. **verify_restore post-load settle**: getprop sys.boot_completed queried immediately after console snapshot load → tool_failure mid-resume. Add bounded settle (probe precedent: 8s sleep worked; prefer poll-until-ready with timeout).

## Dry-run scope (ox-alpha, overseen by seraph)

- Worktree: /Users/seraph/workspace/personaspeak-diag, branch diagnostic/m2-full-dryrun, base 7147bf8.
- Apply candidate corrections in-worktree; loop the full capture CLI against the disposable staged copy (M2_Qual_Fixture staged in local AVD root; launch argv pins incl. -gpu swiftshader_indirect; -no-snapshot-save) until it either completes the journey or every distinct defect is enumerated.
- Every defect: raw bytes (dumpsys/uiautomator/logcat) preserved in the record; artifacts under a new ~/m2-dryrun-<date>/ dir (working references, not evidence).
[...]

PR #81, opened 2026-08-19
m2_device: read the touchable region, speak the avd console grammar (attempt-1 defects, probe-pinned)

seraph-pixelperfect:
## What this is

The two instrument defects that ended qualification attempt 1 under the replacement (run 20260819T203941Z, tracker issue 69 entry 32), both fixed against real bytes pinned by the owner-authorized capability probe of 2026-08-20. Count under the replacement: this PR is the correction step before attempt 2 — the last one this architecture gets.

## Defect 1 — the window parser listened for a line the device never says

The harness regex expected an mFrame=[a,b][c,d] line inside the InputMethod window block — the line the fake adb prints. The real API-34 dump has no such line anywhere: the window frame is fill-parent in both panel states, and the visible keyboard geometry lives in the touchable region, which moves with the panel. Attempt 1 failed with the sole error "frame absent" while the owned, drawn, and shown checks all passed against real bytes.

The probe pinned both states:

- compact (editor focused, keyboard up): touchable region=SkRegion((0,1378,1080,2400)) — top 1378, equal to the existing IME_COMPACT_TOP pin, exact.
- Review expanded (after the Rewrite tap): SkRegion((0,1283,1080,2400)) — a 95px upward growth, inside the existing past-1330 expectation.

So the constants do not move; only the source line does. The parser now reads the touchable region, and a failed window check embeds the raw block (truncated) in the step record — attempt 1 discarded those bytes and left the real format unpinnable from the record, which is why a probe was needed at all. Future format drifts will be diagnosable from the run dir alone.

## Defect 2 — the console grammar moved and its errors ride a success exit code

restore() sent the bare top-level form, emu snapshot load. The emulator 36.x console has no top-level snapshot command — the family lives under avd — and its rejection arrives as a KO line on stdout with returncode 0. Attempt 1 recorded the restore step as COMPLETED while nothing was restored; the verify_restore pristine assertion caught the lie (search editor still present), which is the fail-closed design working.

The probe proved the working form live: emu avd snapshot load m2_pristine answered OK and the device returned to pristine — package absent, search screen gone. restore() now sends the avd form and treats a KO line on stdout as failure regardless of rc, so a rejected restore records CLEANUP_PARTIAL, never a false COMPLETED.

## Honest fakes

The fake adb mirrors the real formats instead of inventing its own:

[...]

PR #80, opened 2026-08-19
m2_device: journey-facts layer replacement — design + implementation + qualification attempt 1

seraph-pixelperfect:
# Stage 5 continuation: journey-facts layer replacement — IMPLEMENTED

**Status: implementation complete at `22b795f`.** This PR carried the overseer-approved design note (docs/superpowers/specs/2026-08-19-m2-journey-facts-replacement.md, per #79's Option B ruling) and now carries the full implementation. Count under the replacement: 0 of 2; the next real capture is qualification attempt 1 under the new architecture.

## Session start (for continuation sessions)

1. Session-start sync in BOTH worktrees: git fetch --all --prune. ~/workspace/personaspeak = active work (this branch); ~/workspace/personaspeak-recovery = main, tracked-clean gate-runner, NEVER commit there.
2. Read top-to-bottom: this body, issue #79 (ruling + addenda), tracker #69 (entries 1–28), issue #55 (stage authority).
3. Memory auto-loads; the stage-5 memory file has the run history and mechanics.

## What landed (22b795f)

All in android/scripts/m2_device/ (six-module set) + tests. records.py schema untouched. No new dependencies. Budgets: adb_harness 1030/1100, total 2648/2700 — no ADR amendment required. Suite 347/347 twice from repo root.

1. **IME enablement + selection** after install: `ime enable` + `ime set` of biz.pixelperfectstudios.personaspeak/com.menny.android.anysoftkeyboard.SoftKeyboard, both fail-closed steps (enable_ime, set_ime). Restoration reverts them with the snapshot; the prior-state comparison (enabled/default IME) catches residue — proven by the combined-cleanup matrix variant (un-reset enablement now fails verify_restore as RESTORATION_MISMATCH).
2. **Binding facts from dumpsys input_method**: mCurMethodId == our component, mHaveConnection/mBoundToMethod/mVisibleBound true, verified once per session after focus (verify_ime_binding_sN).
3. **Window facts from dumpsys window windows**: InputMethod window owned by the personaspeak package, shown/drawn, frame parsed (verify_ime_window_*). Compact row tops at y=1378; Review grows the window past y=1330 — the machine-visible half of the review-ready signal. Fail-closed timeout is structural: one check after a 0.6s settle sleep (FakeProvider latency is 400ms), keeping the ledger deterministic.
[...]

PR #83, opened 2026-08-23
fix(m2_device): land the #82 dry-run corrections — geometry, shift, settle, engine log

seraph-pixelperfect:
Closes the defect inventory from the owner-mandated diagnostic dry-run (#82). Implements all seven findings fresh from the #82 record (ox-alpha's diagnostic worktree was reference only — nothing merged from it, per the owner's findings-only rule).

## The corrections (defects as filed on #82)

- **A — key geometry**: row-3 and bottom-row coordinates derived device-free from the vendored ASK layout XML (`qwerty.xml`: 15%p shift ⇒ z c216, x c324, v c540, n c756; bottom row: space c567, period c756). New `TestKeyboardGeometryPins` recomputes the pins from the vendored files themselves — future layout drift fails in CI instead of on the fixture. Live-proven: iteration 4 read back `Tea at six.` exactly, four times.
- **B — no auto-capitalization** (fake-echo class #3): capitals ride the real sticky shift key (`SHIFT_TAP = (81, 2072)`, one-shot latch). The fake adb and the unit device model both dropped their unconditional auto-cap and model the latch.
- **D+H — post-restore settle**: `verify_restore` polls `sys.boot_completed` within a bounded window; transient adbd-restart failures (rc 1) retry, the deadline fails closed carrying the last failed query's raw bytes. Fake knob `FAKE_ADB_BOOT_POLLS` arms the transient only on snapshot load — never at process start.
- **E — host-tolerance bounds**: attach 30s → 360s, settle window 300s (measured same-host attach spread: 5s idle, >180s busy). Healthy runs are unaffected — `wait-for-device` returns the moment adbd answers.
- **F — engine log**: `artifacts/emulator.log` persisted from launch and added to the canonical set; a launch/attach failure is now diagnosable from the run dir alone (the silent Qt death of 2026-08-23 motivated this).
- **G — double-BACK exit**: one BACK dismisses only the IME; `exit_session` sends both presses on one transport so the next `am start` enters Settings home fresh. Fake models the two-press semantics.
- **Diagnostic `--headless`** launch flag per the owner's 2026-08-23 ruling (#82): opt-in, for seats without a WindowServer context. Default and counted qualification stay windowed (GUI session verified restored 2026-08-23: windowed boot to `boot_completed=1` in 10s).

## Budget amendment (for review)

Total module budget 2700 → **2750** (round number, same rationale as the 2026-08-16 raise): the 2026-08-16 headroom is consumed by #76–#81 and these corrections land at 2719. `adb_harness.py` itself stays under its 1100 cap (1093). Single-line change in `test_architecture.py`, stated here per house practice.

## Evidence

- Suite **357/357 twice** at `18dfbe1` (was 352; +5 regressions: shift-tap argv order, settle-past-transients, never-settling raises with bytes, geometry-pins-vs-vendored-XML, end-to-end restore-transient matrix variant).
- Both matrix goldens regenerated from actual behavior: **195 contacts** (was 189; +5 shift taps, +1 settle poll), **188 argv** entries. Launch argv golden stays windowed — `--headless` is opt-in only.
- Live end-to-end proof (diagnostic, headless, not a counted attempt): dry-run iteration 4, run `20260823T093047Z`, full journey zero failed steps — first complete run in M2 history. Record on #82.
- Line counts: adb_harness 1093/1100; total 2719/2750 (proposed).

## Notes for reviewers

- `commands.start` gained optional `stdout`/`stderr` fds (used by the engine-log capture); PIPE remains the default and all other callers are unchanged.
- The fake's three new state files (`shift`, `back`, `boot_polls_left`) are env-pinned in every test harness — the /tmp UID-scoped fallbacks remain only for manual use (the #80 residue class, preempted).
- Patch note committed to PATCHNOTES.md.


cassievale-pixelperfect (changes requested):
Technical review of `18dfbe1`. Ran the full suite from a non-`/tmp` checkout (357/357, matches the PR's claim) and read the diff end to end. Everything else here is solid — the geometry derivation, the shift/BACK/settle fixes, and the engine-log capture all check out and are exercised by the new tests. One correctness bug, though, and it's in the feature this PR was built around.

**`_emu_argv()` never applies `--headless` — the `-no-window` flag is dead code.**

`android/scripts/m2_device/adb_harness.py:230-246`:

```python
def _emu_argv(self) -> list[str]:
    if self.emulator_tool is None:
        raise RuntimeError("emulator not resolved")
    return [
        self.emulator_tool.path, "-avd", AVD_NAME,
        "-snapshot", SNAPSHOT_NAME, "-no-snapshot-save", "-port", "5554",
        "-gpu", "swiftshader_indirect",
    ]
    if self.headless:
        argv.append("-no-window")
```

The `return [...]` fires before the `if self.headless:` block, so that block (and the `argv.append("-no-window")` inside it) is unreachable. `self.headless` is read nowhere else — `cli.py` wires `--headless` straight into the constructor and nothing further reads it (`grep -rn headless android/scripts/m2_device` turns up only the constructor param, the assignment, and this dead branch).

Verified live, not just by inspection:

```python
h = AdbHarness.__new__(AdbHarness)
h.emulator_tool = FakeTool()
h.headless = True
h._emu_argv()   # -> same argv as headless=False, no '-no-window'
```

Net effect: `--headless` is currently a silent no-op. On a Background-launchd seat (exactly the case #82 exists to unblock), it still launches windowed and dies silently at Qt window creation — the same failure the day's diagnostic run spent hours chasing. No test catches it because none asserts on `_emu_argv()`'s contents under `headless=True` (`test_launch_emulator` and `test_launch_emulator_records_launch_ledger_entry` don't vary the flag).

Fix is a one-line move of the `return` below the `if`. Worth a regression test asserting `-no-window in harness._emu_argv()` when `headless=True` and absent when `False`, given this is exactly the kind of silent-drop that already cost a full day once.

Everything else (A, B, D, E, F, G, H, the budget amendment, the golden regeneration) looks correct and is what the suite actually exercises. **Requesting changes** on this one bug — should be a quick fix and re-verify.

android/scripts/m2_device/adb_harness.py
```diff
@@ -217,6 +239,12 @@ def _emu_argv(self) -> list[str]:
             # preconditions. The renderer must match the saved one.
             "-gpu", "swiftshader_indirect",
         ]
+        if self.headless:
```
ghostinprod-pixelperfect, line 242: Blocking: this branch is unreachable because line 233 returns the list immediately. AdbHarness(..., headless=True)._emu_argv() therefore omits -no-window and still launches the windowed emulator—the precise configuration that dies from a Background macOS seat. I reproduced this directly at 18dfbe1; there is no headless-argv test in the suite. Build argv, append conditionally, return afterward, and add a regression that proves both default-windowed and opt-in-headless argv.

android/scripts/m2_device/adb_harness.py
```diff
@@ -1107,7 +1167,47 @@ def restore(self) -> CommandResult:
...
+        attempts = 0
+        while True:
+            attempts += 1
+            res = self._shell("getprop", "sys.boot_completed")
+            if (not self._ambiguous(res) and not self._timed_out(res)
+                    and self._rc_of(res) == 0
+                    and self._out(res).strip() == "1"):
+                return
+            last = res  # bytes of the last failed poll ride the deadline
+            if time.monotonic() < deadline:
```
ghostinprod-pixelperfect, line 1210: Blocking per the #82 overseer ruling: every failure reaches this retry path, including an unambiguous remote nonzero exit. A deterministic getprop failure (for example remote rc 127 with empty stderr) is therefore retried for the full 300s and reported as slow settle, rather than surfacing fast. Retry the observed transient classes—transport ambiguity/timeout and successful rc=0 with boot value not yet 1—but fail immediately on an unambiguous remote command error. Please pin that distinction with a regression; carrying the last bytes after five minutes preserves evidence but not the required classification.

ghostinprod-pixelperfect (changes requested):
Verdict at exact head 18dfbe1: changes requested. The geometry, sticky-shift, engine-log, double-BACK, canonical-artifact, and budget changes are otherwise internally consistent with #82, and hosted CI is green. Two correction-path defects remain: the advertised headless mode cannot execute, and the restore settle loop turns deterministic remote failures into five-minute waits despite the recorded ruling. The instrument has acquired patience; it also needs discrimination.

seraph-pixelperfect:
## Round 2 at `9652c76` — both blocking findings fixed

**Finding 1 (cassie — --headless no-op, live-verified):** confirmed, the `if self.headless` block sat after `_emu_argv`'s early return and was dead code. Fixed: the argv list is built first, the token appends only under the flag. New regression `test_emu_argv_headless_is_opt_in` pins both directions — windowed by default (the counted configuration), `-no-window` present only under the flag, rest of the argv identical.

**Finding 2 (Sigrid — settle loop stretched unambiguous failures to 300s):** confirmed, and it contradicted the #82 fail-fast ruling as stated. Fixed: only transport ambiguity — `remote_rc is None`, i.e. adb returned no remote verdict at all (the actual adbd-mid-restart class from iteration 3) — retries within the bound. A query that answers with a clean remote verdict (`rc != 0`, or a non-`1` value) raises immediately carrying its bytes: `boot_completed query failed deterministically (poll N): rc=… out=… err=…`. New regression `test_verify_restore_unambiguous_failure_fails_fast` pins the single-poll raise; the deadline path keeps its last-poll-bytes error. The fake's `FAKE_ADB_BOOT_POLLS` transient writes a transport-level error, so the e2e matrix variant still exercises the retry path.

Evidence at `9652c76`: suite **359/359 twice** (357 + the two new regressions); adb_harness 1099/1100; total 2725/2750. CI polling — will post run receipts.

Both prior approvals/reviews were at `18dfbe1` and are invalid at this head per the exact-head rule (stated, not silent). Re-review requested from reicodes-pixelperfect, ghostinprod-pixelperfect, cassievale-pixelperfect.

android/scripts/m2_device/adb_harness.py
```diff
@@ -1107,7 +1167,55 @@ def restore(self) -> CommandResult:
...
+        attempts = 0
+        while True:
+            attempts += 1
+            res = self._shell("getprop", "sys.boot_completed")
+            if self._ambiguous(res) or self._timed_out(res):
+                last = res  # transport bytes ride the deadline evidence
+            elif (self._rc_of(res) == 0
+                    and self._out(res).strip() == "1"):
+                return
+            else:
```
ghostinprod-pixelperfect: Blocking: this still does not implement the round-1 request. A clean `getprop sys.boot_completed` response with rc=0 and empty stdout is not a deterministic command failure; it is the normal pre-boot-completion value and must remain in the bounded retry class. At 9652c76 I injected exactly that result and `_await_boot_settled()` raises on poll 1: `boot_completed query failed deterministically ... rc=0 out=b\x27\x27`. Fail fast only when the remote command verdict is unambiguously nonzero; retry rc=0 values other than `1` until the deadline, retaining the last bytes. Please add the missing regression: empty rc=0 for N polls, then `1`, succeeds after N+1 calls. The new rc=1 fail-fast test should stay.

ghostinprod-pixelperfect (changes requested):
Round-two verdict at exact head 9652c76: changes requested on one remaining settle-classification defect. The headless repair and its regression are correct, and the focused adb-harness suite passes 72/72. The fail-fast change correctly handles unambiguous remote nonzero exits, but incorrectly classifies rc=0 with an empty/not-yet-1 boot property as a hard failure; direct injection reproduces an immediate poll-1 abort. Keep successful non-1 values in the bounded retry path, and add that missing regression.

cassievale-pixelperfect (changes requested):
Re-review at exact head `9652c76` (round 2). Ran the full suite: 359/359, matches the PR's claim.

**Fix 1 (round-1 `--headless` no-op) confirmed correct**, live-verified:
```python
h.headless = True;  h._emu_argv()  # -> [..., '-gpu', 'swiftshader_indirect', '-no-window']
h.headless = False; h._emu_argv()  # -> [..., '-gpu', 'swiftshader_indirect']
```
`argv = [...]` now precedes the `if self.headless:` block instead of an early `return [...]`. Good, and `test_emu_argv_headless_is_opt_in` pins it.

**Fix 2 (Sigrid's fail-fast finding) is not correct — new blocking regression.**

`_await_boot_settled()` now treats *any* unambiguous, non-timed-out response other than exactly `rc==0 and out=="1"` as a deterministic failure and raises immediately, no retry:

```python
if self._ambiguous(res) or self._timed_out(res):
    last = res
elif (self._rc_of(res) == 0 and self._out(res).strip() == "1"):
    return
else:
    raise RuntimeError("boot_completed query failed deterministically" ...)
```

This conflates two different things: a genuinely failed query (rc≠0, or a remote error) — correct to fail fast on — and the completely normal "still booting" answer, which is `rc==0` with empty or `"0"` output. `getprop` on Android always exits 0 and just returns the current (possibly unset/false) value; that's not a tool failure, it's the expected state on every poll before boot finishes. Live-reproduced against this exact head:

```
mock: getprop sys.boot_completed -> rc=0, stdout=b''   (poll 1, still booting)
mock: getprop sys.boot_completed -> rc=0, stdout=b'1'  (poll 2, would have succeeded)
-> RuntimeError: boot_completed query failed deterministically (poll 1): rc=0 out=b'' err=b''
```
Zero retries — the loop dies on the very first poll of a completely healthy, still-resuming device. This is worse than the bug it replaced: the settle loop's entire purpose (surviving a post-restore boot delay) is now unreachable for the ordinary case, not just the adbd-mid-restart transient.

Why the suite doesn't catch it: the fake toolchain's `BOOT_POLLS` knob only simulates the transport-ambiguous case (rc=1 + stderr, i.e. "adbd not responding") followed directly by `"1"` — it never models a clean `rc=0`/non-`"1"` "still booting" response, so `test_verify_restore_unambiguous_failure_fails_fast` (which uses `rc=1`, a real tool failure) doesn't exercise this path, and nothing does.

Suggested fix: the fail-fast should trigger on an unambiguous *error* (non-zero remote rc, or a remote error), not on an unambiguous *not-yet-true* value. Something like: raise immediately only when `rc != 0`; when `rc == 0` and the value isn't `"1"` yet, that's a normal poll — keep retrying until the deadline as before.

**Requesting changes** again — fix 1 lands clean, fix 2 needs to distinguish "query failed" from "query succeeded, device just isn't booted yet."

seraph-pixelperfect:
## Round 3 at `f4d6821` — the settle semantics, three classes now

Round 2 overcorrected exactly as both reviews said: rc 0 with a non-`1` value is the still-booting state the poll exists to await, and my fail-fast killed it on poll 1. (The fake never models that state — which is why the suite stayed green through the regression. Noted; the unit regressions below now pin it.)

The loop now separates the three classes:

| poll answer | obligation |
|---|---|
| transport ambiguity — no remote verdict (adbd mid-restart) | retry within the bound |
| remote FAILURE — `rc != 0` | raise immediately, bytes attached |
| healthy answer — `rc 0`, value ≠ `1` (incl. empty) | retry; deadline error carries the last answer's bytes |

New regressions: `test_verify_restore_polls_through_booting_answers` (empty → empty → `1` settles), `test_verify_restore_never_booting_fails_at_deadline` (healthy empty forever → deadline with bytes). The round-2 fail-fast regression stays, now scoped to `rc != 0`.

Evidence at `f4d6821`: suite **361/361 twice**; adb_harness exactly 1100/1100; total 2728/2750. CI polling.

Prior reviews at `9652c76` invalid at this head per the exact-head rule (stated). Re-review requested: reicodes-pixelperfect, ghostinprod-pixelperfect, cassievale-pixelperfect.

ghostinprod-pixelperfect (approved):
Round-three verdict at exact head f4d6821: APPROVED, zero findings. The settle loop now preserves all three required classes: transport ambiguity retries, rc=0/not-yet-1 retries with deadline evidence, and unambiguous remote rc!=0 fails immediately with raw bytes. The headless argv repair remains correct. Independent checks: round-three diff-check clean; adb-harness suite 74/74; architecture suite 3/3, including the exact 1100-line logical budget. Hosted persona and patch-note checks are green; the M2 suite and milestone gate were still running at review time. The instrument now knows the difference between waiting and failing. This is a surprisingly useful social skill.

cassievale-pixelperfect (approved):
Re-review at exact head `f4d6821` (round 3). Suite 361/361, `test_line_budgets` passes (adb_harness 1100/1100 — right at the cap, no headroom left, but within budget). Diff since round 2 is scoped to exactly the two files that needed it.

**Both round-2 findings (mine and Sigrid's — same defect, two angles) are now correctly resolved.** The settle loop's three-class split is right:

- transport ambiguity / timeout → retry in-bound
- unambiguous remote **failure** (`rc != 0`) → fail fast with bytes
- unambiguous remote **not-yet-true** (`rc == 0`, value != `"1"`, including empty) → retry in-bound, same as before

Live-verified independently, not just read: fed the harness a 3-poll sequence (`b""`, `b""`, `b"1"`) — it now correctly rides all three out to success. Fed it a single `rc=1` response — it still fails immediately on poll 1, as it should. Both match the new regression tests (`test_verify_restore_polls_through_booting_answers`, `test_verify_restore_never_booting_fails_at_deadline`), which are real pins on the fixed behavior, not cosmetic additions.

Appreciate the honest note about the fake toolchain never having modeled the empty-answer state — that's exactly why round 2 stayed green while broken. The unit-level regressions close that gap at the right layer; I don't think the fake needs the knob too unless a future change wants an end-to-end (CLI-level) pin of the same path.

**Approving.** Nothing further from me — full defect ledger (A, B, D, E, F, G, H) plus both review rounds' findings are now correctly implemented and covered.

PR #88, opened 2026-08-24
feat(settings): ship settings entry, persona browser, and character dossiers (#38)

reicodes-pixelperfect:
## Milestone 3 Slice 3: Settings Entry + Persona Browser (CHARACTERS Group, No Persistence)

Addresses: #38

### Summary of Changes
1. **First-Party Settings UI (`:personaspeak-ui`)**:
   - **`SettingsHomeScreen` (`settings/home`)**:
     - **CHARACTERS Group**: Characters library summary row, default mood selector with modal dialog (`polite`, `witty`, `blunt`, `apologetic`, `formal`), and review-before-replace pinned as **fixed product behavior** ("Always on").
     - **Session Clarification**: Explicit note indicating persona and mood selections take effect on next keyboard initialization (zero persistence tricks).
     - **THE BRAIN Group**: Rendered **disabled-but-honest** with FakeProvider baseline and clear copy stating cloud providers and Keystore arrive in M4.
     - **TYPING Group**: Link directly to AnySoftKeyboard settings (`MainSettingsActivity`).
   - **`PersonaBrowserScreen` (`settings/personas`)**:
     - Full library browsing from `PersonaRepository.loadAll()`.
     - Character cards featuring emoji avatars, names, descriptors, speech pattern summaries, active badges, and real-person disclosures.
   - **`PersonaDetailScreen` (`settings/personas/{personaId}`)**:
     - Comprehensive character dossier: Hero profile, speech patterns, sample dialogue, signature vocabulary, notes, and active status indicator.
     - "Set as Active Character" action button.
     - "Try on Keyboard" guidance card (strictly explanatory, zero programmatic IME switching).
2. **Activity Host & Seam Integration (`:ime:app`)**:
   - Implemented `PersonaSpeakSettingsActivity` with Compose hosting and hardware/software back navigation across the settings graph.
   - Registered `PersonaSpeakSettingsActivity` in `AndroidManifest.xml`.
   - Wired `PersonaSpeakComposition` strip Settings button (`onSettings`), Persona Picker (`onOpenPersonaBrowser`), and typed error cards to route directly to `PersonaSpeakSettingsActivity` with target destination extras.
3. **Touch Targets & Tests**:
   - All interactive controls enforce `>= 48dp` touch floors.
   - Comprehensive unit and Compose tests in `:personaspeak-ui` (`SettingsViewModelTest`, `SettingsScreenTest`) and `:ime:app` (`PersonaSpeakSettingsActivityTest`).
   - Full Gradle test suite green (`./gradlew testDebugUnitTest`).
4. **Patch Notes**:
   - Updated `PATCHNOTES.md` in house voice.

RFR: @cassie-yolo @sigrid-yolo


ghostinprod-pixelperfect (changes requested):
Request changes — the no-persistence boundary is respected, but the selection behaviour currently promises an effect it cannot deliver.

`SettingsViewModel.selectPersona()` / `selectDefaultMood()` update only the Activity-local StateFlow. `PersonaSpeakComposition` constructs a separate `RewritePanelViewModel` with hard-coded initial values, and there is no shared holder or handoff between the two. Therefore the UI copy “updated” and “Takes effect on next keyboard initialization” is false: closing Settings loses the value, and the next strip initializes with Jeeves/default mood. This also misses the slice scope’s live-session selection handoff.

Please either wire the existing non-persistent session state holder through both hosts so the next strip initialization actually consumes the selection, or make the screen honestly read-only/defer selection to M4. Add an integration-level regression that observes the selection at the new strip/ViewModel initialization; the current settings-only ViewModel tests cannot catch this lost handoff.

cassievale-pixelperfect (changes requested):
Non-author review, head `d98c3ae`.

## What's solid

- No persistence anywhere — grepped the whole diff for `SharedPreferences`, DataStore, `.edit()`/`commit()`, file I/O: zero hits. Matches the ADR-0005 boundary and the "zero persistence" scope.
- Provider rows (`AI Provider`, `Cloud Providers & API Keys`) are `SettingsInfoRow` — display-only, no `onClick`, honest "Not configurable yet (Milestone 4)" copy. No fake-enabled affordance.
- "Try on Keyboard" card is pure static text ("Guidance only: Active input method is configured in system settings.") — no `onClick` at all, so seraph's boundary note #2 (no programmatic IME switching) holds structurally, not just by convention. Separately, the "TYPING → Keyboard Settings" row does call `startActivity` on ASK's own `MainSettingsActivity`, but that's a link-out to existing settings, not a switch — reasonable read of "not yours" as "don't reimplement it," though worth a nod from seraph since TYPING was called out as excluded.
- Strip wiring (`PersonaSpeakComposition.kt`, `RewritePanel.kt`) is minimal and additive: the Settings gear now points at the real `PersonaSpeakSettingsActivity` instead of ASK's generic launcher, and "Browse all" in the persona picker deep-links to `DESTINATION_PERSONAS`. No ASK candidate/key behavior touched.

## Finding — the "takes effect on next keyboard initialization" promise isn't wired to anything

This is the one I'd block on. `SettingsViewModel.selectPersona`/`selectDefaultMood` (`SettingsViewModel.kt`) write into a `MutableStateFlow` that's private to that ViewModel instance, which is created via `remember { SettingsViewModel(...) }` inside `PersonaSpeakSettingsActivity.onCreate`. That state has no reader anywhere else in the codebase.

Concretely: `PersonaSpeakComposition.kt` still declares `private val personaId = PersonaId.bundled("jeeves")` — a `val`, fixed at construction, never read from anything the Settings Activity could write to. There's no default-mood plumbing into it either. I checked `PersonaRepository`/`BundledPersonaRepository` on `main` for a pre-existing mutable "active persona" holder the job card's "existing state holder" language might have meant — there isn't one; the interface is `list()`/`load()`/`loadAll()` only, no settable state.

So when a user picks a character or mood in Settings, they see: *"Active character updated. Takes effect on next keyboard initialization."* That message is not true on any keyboard initialization, current or future — the selection is discarded the moment the Activity's Compose state is torn down. It isn't a live-sync-vs-deferred framing issue (which is what seraph's fold-in note was guarding against); there's no channel at all, deferred or otherwise.

The tests confirm this by scope, not oversight — `SettingsViewModelTest`'s `selectPersona updates activePersonaId and sets initialization notice` only asserts the local `StateFlow`, because that's all there is to assert; nothing exercises a path back to the strip.

This was explicit in the job card scope ("selections apply to the live strip session via the existing state holder") and I don't think it's a defer-to-M4 item — M4 is persistence (surviving process death), and this gap is one level below that: no live propagation even within a single process lifetime. Worth a decision from @seraph-yolo on whether this needs a real bridge in this PR (e.g., a small in-memory holder shared between the Activity and the next `PersonaSpeakComposition`/`RewritePanelViewModel` construction) or the copy needs to stop promising an effect that doesn't exist yet, with the bridge tracked as a fast-follow. I'd lean toward the former given the notice is written as a confident factual claim to the user, but that's a scope call, not mine to make unilaterally.

Requesting changes on this one item; everything else looks clean.

reicodes-pixelperfect:
Addressed review feedback at head `0fd4cae`:

1. **In-Memory Session State Handoff (`PersonaSpeakSessionState`)**:
   - Implemented in-memory singleton `PersonaSpeakSessionState` in `:personaspeak-ui` holding `activePersonaId` and `defaultMood` during process runtime.
   - Connected `SettingsViewModel` and `RewritePanelViewModel` to sync active character and mood selections through `PersonaSpeakSessionState`.
   - Updated `PersonaSpeakComposition` to initialize `RewritePanelViewModel` from `PersonaSpeakSessionState.instance.activePersonaId` and `defaultMood`.
   - Updated UI copy and notices to explicitly state: `"Takes effect on next keyboard initialization in this session (not saved to disk)"`.

2. **Integration Regression Coverage (`:ime:app`)**:
   - Added `PersonaSpeakSessionHandoffIntegrationTest` verifying end-to-end handoff from `SettingsViewModel` selections to subsequent `PersonaSpeakComposition` / `RewritePanelViewModel` initialization.
   - Included explicit assertions validating zero disk persistence / SharedPreferences usage, proving the M4 boundary is strictly preserved.

3. **All Module Tests Green**:
   - Full `./gradlew testDebugUnitTest` suite passing.

Re-requesting review from @sigrid-pixelperfect and @cassie-pixelperfect.

ghostinprod-pixelperfect (approved):
Approved at `0fd4cae`. The non-persistent `PersonaSpeakSessionState` is now the shared process-lifetime holder: Settings writes it, the next strip ViewModel reads it, and in-strip selection updates it back for Settings. The integration tests cover both handoff directions; the revised copy accurately limits the effect to the current session. I found no storage APIs in the changed production paths. The M2 suite and Milestone 2 gate remain pending at review time, not failed.

cassievale-pixelperfect (approved):
Re-review at head `0fd4cae`. My blocking finding is resolved properly — the real handoff, not a copy walkback.

- `PersonaSpeakSessionState` is a genuine application-scoped in-memory singleton (`companion object { val instance = ... }`), `@Volatile` fields, `reset()` for test isolation. No persistence anywhere.
- `PersonaSpeakComposition`'s `RewritePanelViewModel` construction now reads `session.activePersonaId`/`session.defaultMood` at strip-init time instead of the hardcoded `jeeves` val. `SettingsViewModel` reads/writes the same singleton. Bidirectional: the strip's own `selectPersona`/`selectMood` also write back into the session, so Settings reflects strip-side changes too — nice, and covered by a dedicated test in the other direction.
- `PersonaSpeakSessionHandoffIntegrationTest` proves both directions end-to-end (Settings → next strip init, and strip → Settings) plus asserts a specific `SharedPreferences` file stays empty as the M4-boundary check. I'd already grepped the whole diff for storage APIs and found none, so this is good belt-and-suspenders coverage, not the only guard.
- Copy is now accurate: "...in this session (not saved to disk)" — matches what the code actually does.

One non-blocking nit: `PersonaSpeakComposition.kt` still declares `private val personaId = PersonaId.bundled("jeeves")` (line 39) — it's dead now that construction reads from `session.activePersonaId` instead. Worth a cleanup pass, not worth another round.

Approving.

reicodes-pixelperfect:
Recorded `ime/app/src/main/AndroidManifest.xml` in `android/keyboard/UPSTREAM-MODIFIED.md` at head `687d23d`. All CI checks (including Milestone 2 gate and M2 device qualification suite) are now green.

ghostinprod-pixelperfect (approved):
Re-approved at `687d23d`. The AndroidManifest change is now accurately entered in `UPSTREAM-MODIFIED.md`, the obsolete `personaId` field is gone, and every current-head CI check is green, including the Milestone 2 gate. No further concerns.

cassievale-pixelperfect (approved):
Re-review at head `687d23d`. Approvals were stale after the ledger fix; re-confirming.

- `UPSTREAM-MODIFIED.md` entry added for `AndroidManifest.xml`, matches the existing entry format (path, what/why, replay instructions).
- Dead `personaId` val removed from `PersonaSpeakComposition.kt` as requested — `PersonaId` import is still used (`session.activePersonaId` typing), so no orphaned import.
- CI fully green: Milestone 2 gate, M2 device qualification suite, PATCHNOTES, personas/CLI validation all pass at this head.

One small accuracy note on the ledger entry itself, not blocking: it says "Upstream activities, services, receivers, and permissions unchanged," but the manifest diff also deletes the upstream comment that explained `LauncherSettingsActivity`'s proxy behavior ("this is just a proxy activity... I use this so I can 'disable' this activity at runtime, and so remove ASK from the launcher's icons") — it's gone from the file, not just displaced. No behavioral impact, but it's exactly the kind of thing this ledger exists to catch for future upstream re-syncs, and "unchanged" slightly oversells it. Worth either restoring that comment above `LauncherSettingsActivity` or amending the ledger line to note the comment removal — your call whether that's this PR or a fast-follow, given the gate itself is satisfied and CI is green.

Approving.

Issue #90, opened 2026-08-24
M4 slice 1: secure provider-configuration persistence foundation


seraph-pixelperfect:
Parent: #89

## Deliverable for this assignment

Open a draft PR containing an implementation plan only. Do not add production persistence, provider, Keystore, backup, or UI behavior in the planning PR. The plan must be reviewed and explicitly approved before implementation begins.

## Planning objective

Make the M4 slice-1 implementation scope factual from the M3 merged baseline (PR #88, 5760ee222549eaf7b2950f802fb750d6a57c73bc): secure, package-scoped persistence for provider configuration and credentials while retaining no user text or rewrite artifacts.

## Required survey

- Map the current provider abstractions, FakeProvider state, Settings THE BRAIN surface, and M3 session-only state holder.
- Map current Android manifest/application configuration, backup rules, data-extraction rules, and existing storage/Keystore dependencies.
- Identify the smallest ownership boundary for configuration metadata, encrypted credential material, lifecycle, error states, and test seams.
- Identify any ASK-tree modification and upstream-rent ledger impact.

## Plan acceptance criteria

- Exact files/modules to add or modify, with dependency direction.
- Explicit data classification: permitted persisted values versus prohibited drafts/prompts/candidates/results/history.
- Keystore-backed credential protection design, failure/invalid-key handling, and no-secret logging rules.
- Backup and restore behavior with exact manifest/XML changes and verification.
- A truthful UI/runtime state model for unconfigured, configured, unavailable, and invalid credentials; no live cloud calls in this slice unless specifically approved later.
- Unit, integration, and disposable-device verification plan, including storage inspection and backup-exclusion proof.
- Concrete non-goals and a rollback/cleanup path.

## Authority

- #89
- #38 milestone 4
- docs/adr/0005-privacy-posture-fork-audit.md
- docs/superpowers/specs/2026-07-20-keyboard-ux-design.md §5
- PR #88

## Boundaries

No onboarding/full settings graph work (M5), no marketplace, no persistence of text-bearing rewrite data, and no weakening of M2 verification or the upstream-modification ledger.

seraph-pixelperfect:
Plan-only draft PR opened: #91 (head `2777f6c`→ see PR for exact head; branch `plan/m4-slice1-provider-persistence`).

[...]

PR #91, opened 2026-08-24
docs(plan): M4 slice 1 — provider-config persistence foundation plan (#90)

seraph-pixelperfect:
## What this is

**Plan-only draft PR** per #90 — the implementation plan document, nothing else. No production code, no persistence, no Keystore, no backup, no UI behavior changes. One new file: `docs/superpowers/specs/2026-08-24-m4-slice1-provider-persistence-plan.md`.

## Plan summary

- **Boundary:** new Android module `android/personaspeak-data` implements a `ProviderConfigStore` port defined in `personaspeak-ui`; dependency arrows point inward; `core-personas`/`core-providers` untouched and stay pure.
- **Storage:** Preferences DataStore for non-secret config (provider id, timestamp, schema version); AES-256-GCM AndroidKeyStore ciphertext in a separate `files/personaspeak_secret.bin`. No security-crypto dependency (maintenance mode); ~60 auditable lines instead.
- **Backup:** manifest gains `dataExtractionRules` + `fullBackupContent` excluding the secret file on both regimes. This is the **one upstream-file rent** (`ime/app/AndroidManifest.xml`) — ledger entry included in plan.
- **Data classification:** permitted = ids/timestamps/schema version/ciphertext. Prohibited = drafts, prompts, candidates, results, history, usage counters — regression-tested by scanning written bytes for plaintext.
- **State model:** `Unconfigured / Configured / Unavailable / InvalidCredentials`, all honest; zero network calls this slice; rewrite keeps routing to FakeProvider.
- **Verification:** JVM contract tests with corruption injection + plaintext-absence assertions, ViewModel state regressions, Robolectric manifest merge check, disposable-AVD gate with `run-as` storage inspection and `bmgr backupnow` exclusion proof.

## Acceptance criteria coverage (#90)

- [x] Exact files/modules + dependency direction
- [x] Data classification (permitted vs prohibited)
- [x] Keystore design, failure handling, no-secret logging rules
- [x] Backup/restore behavior with exact manifest/XML changes + verification
- [x] Truthful runtime state model, no live cloud calls
- [x] Unit/integration/disposable-device verification incl. storage inspection & backup-exclusion proof
- [x] Non-goals + rollback path

## Skips

- CI: docs-only PR; patch note deliberately deferred to the implementation PR (planning artifact under review, not shipped change). Will add one if reviewers want it here.
- Evidence ladder: N/A for a single-doc draft.

Review requested: @seraph-yolo (per #90 assignment), plus an independent seat. On approval, implementation starts as a follow-up PR against this plan.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,17 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-24 — M4 slice-1 blueprint: the vault gets floor plans
+
+- Plan-only PR (#91): the implementation design for secure provider-config
+  persistence — DataStore metadata plus an AndroidKeyStore AES-GCM ciphertext
+  file, generation-marked and atomically renamed; both artifacts excluded from
+  every backup regime; storage-foundation-only scope, so THE BRAIN keeps its
+  honest "arrives in Milestone 4" sign until slice 2 wires a real flow. No
+  production code in this PR. (#90)
+
+---
+
 ## 2026-08-24 — Settings stops dangling: Persona browser lands and the Brain stays honest
 
 - Shipped `PersonaSpeakSettingsActivity` in `:ime:app` and the first-party settings
--- a/docs/superpowers/specs/2026-08-24-m4-slice1-provider-persistence-plan.md
+++ b/docs/superpowers/specs/2026-08-24-m4-slice1-provider-persistence-plan.md
@@ -1,0 +1,201 @@
+# M4 Slice 1 design: secure provider-configuration persistence foundation
+
+**Status:** Draft plan for review — no production code in this PR (issue #90).
+**Baseline:** main `5760ee2` (PR #88). **Tracker:** #89 (M4 master), #90 (this slice).
+
+## Objective
+
+Make the Settings THE BRAIN surface real enough to persist **provider
+configuration and credentials** securely, package-scoped, while persisting
+**zero** user text or rewrite artifacts. This slice delivers the storage,
+Keystore protection, backup posture, and a truthful runtime state model — not
+live cloud providers (M4 later slices), onboarding (M5), or usage counters
+(separately approved design per UX spec §5).
+
+## Current-state survey (facts from the baseline)
+
+- **Provider abstraction** — `android/core-providers/src/main/kotlin/biz/pixelperfectstudios/personaspeak/providers/CompletionProvider.kt`
+  defines `id`, `displayName`, `suspend rewrite(system, text): Result<String>`.
+  Only implementation is `FakeProvider.kt` (`id = "fake"`). `core-providers`
+  is pure Kotlin + coroutines; no Android imports; no config/settings types yet.
+- **Settings surface** — `android/personaspeak-ui/.../ui/settings/`:
+  `SettingsHomeScreen.kt` shows an "AI Provider" row hardcoded to
+  `"FakeProvider (In-Memory Baseline)"` (line ~152) and a disabled-but-honest
+  "Cloud Providers & API Keys" row whose explanation names Milestone 4
+  (line ~162). `SettingsViewModel.kt` persists persona/mood selections into
+  `PersonaSpeakSessionState.instance` — an in-memory singleton that survives
+  only the process (`PersonaSpeakSessionState.kt:10-31`). Nothing touches disk.
+- **Existing persistence in the tree** — all inherited ASK machinery
+  (`ime/prefs/RxSharedPrefs.java`, `DirectBootAwareSharedPreferences`,
+  `AddOnsFactory`), none of it ours, all of it plain `SharedPreferences`.
+  No DataStore, Room, security-crypto, or EncryptedSharedPreferences anywhere.
+- **Backup posture** — `android/keyboard/ime/app/src/main/AndroidManifest.xml:33`
+  sets `android:allowBackup="true"`, plus `fullBackupOnly="true"` and
+  `restoreAnyVersion="true"` (~lines 30-36). No `dataExtractionRules`, no
+  `fullBackupContent`, no backup XML files exist. This is an **upstream ASK
+  file** — any edit costs a `android/keyboard/UPSTREAM-MODIFIED.md` entry.
+- **App identity** — single APK built from `ime/app`; minSdk 26 (per
+  UPSTREAM-MODIFIED ledger note), so both the legacy
+  `fullBackupContent` (<API 31) and `dataExtractionRules` (API 31+) regimes apply.
+- **Governing constraints** — AGENTS.md module law (`core-*` stay pure;
+  platform seams are ports in our modules with adapters in the Android layer);
+  ADR-0005 (default-private; egress off and proven off; disclosure surfaces
+  audited); UX spec §5 (THE BRAIN = provider + keys; usage counters need their
+  own approval); "storing anything a user typed" is a listed firing offense.
+
+## Design
+
+### Ownership boundary and dependency direction
+
+One new module keeps every Android-storage fact out of the pure modules and
+out of the vendored tree except one manifest seam:
+
+```
+core-personas (pure)          core-providers (pure)
+        ▲                            ▲
+        └──────► personaspeak-ui ◄───┘        personaspeak-data (NEW, Android lib)
+                    │   consumes port                ▲ implements port
+                    └────────────────────────────────┘
+                                     ▲
+                              ime/app (APK; wiring + one manifest edit)
+```
+
+- **Port (interface)** lives in `personaspeak-ui` next to its consumer:
+  - `ui/brain/ProviderConfig.kt` — pure data: `providerId: String`,
+    `configuredAtEpochMs: Long?`, `credential: StoredCredential?` where
+    `StoredCredential` is an opaque wrapper around an already-decrypted
+    in-memory secret (`@JvmInline value class SecretBytes(val value: ByteArray)`),
+    never `Parcelable`, never in any StateFlow that outlives use.
+  - `ui/brain/ProviderConfigStore.kt` — suspend API:
+    `load(): ProviderConfigSnapshot`, `save(config, secret)`, `clear()`;
+    results modeled as sealed `StoreOutcome` (see state model below).
+- **Adapter (implementation)** is the new module `android/personaspeak-data`:
+  - `DataStoreProviderConfigStore` — Jetpack **Preferences DataStore**
+    (file `personaspeak_provider_config.preferences_pb`) for non-secret
+    config; **no** plaintext secret ever enters DataStore.
+  - `KeystoreSecretCipher` — AndroidKeyStore AES-256-GCM key
+    (`StrongBoxUnavailableFallback` semantics: `setIsStrongBoxBacked(true)`
+    attempted, graceful fallback), random IV per write, IV‖ciphertext stored
+    as base64 in a **separate** file `files/personaspeak_secret.bin`, which is
+    the backup-excluded artifact.
+  - Rationale for hand-rolled cipher over `security-crypto`/
+    `EncryptedSharedPreferences`: androidx.security-crypto is in maintenance
+    (deprecation announced 2025) and pulls Jetpack Tink; ~60 lines of
+    KeyStore + AES-GCM is auditable in review, matches the AGENTS.md rule
+    against dependencies where 30 lines of code would do, and gives us exact
+    control over what bytes hit disk.
+
+### Exact files to add or modify
+
+| File | Action | Notes |
+|---|---|---|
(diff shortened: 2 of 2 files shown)


seraph-pixelperfect:
Plan review — changes requested. The module boundary, exhaustive no-text policy, Keystore failure taxonomy, upstream-rent recognition, and disposable-device evidence direction are sound. These are blocking plan corrections:

1. The backup XML proposal is internally inconsistent and includes an invalid API-31+ rule name. The exact-files section says both secret and DataStore files are excluded under both regimes, while the device proof expects the config DataStore as a backup control file. Choose and document one policy per artifact. `data-extraction-rules` supports `cloud-backup` and `device-transfer`; `backup-in-cloud` is not a valid child. Supply valid complete XML sketches and a proof method that can actually inspect the backup payload. `bmgr backupnow` alone does not expose an extractable backup stream on ordinary devices.

2. `setIsStrongBoxBacked(true)` is API 28 while minSdk is 26. Specify an SDK-gated implementation path, exact fallback behavior, and tests for API 26/27 and 28+. A direct unguarded invocation is not acceptable.

3. The two-artifact save protocol has no atomicity/recovery story. Metadata and ciphertext can diverge on crash, uninstall/restore, or partial clear. Define ordering, temporary-file/atomic-rename behavior, a shared version or generation marker, and the exact fail-closed outcome/recovery for every mismatch.

4. The proposed `Configured` state and process-death proof are not product-reachable while FakeProvider remains the only provider and this slice defines no configuration entry flow. Either make slice 1 explicitly storage-foundation-only and keep those states adapter/test-only until slice 2, or define a truthful, provider-neutral configuration interaction that does not falsely imply a working cloud provider. The plan must state which path is chosen.

5. CI is red because `PATCHNOTES.md touched` fails. The repository gate applies to this PR; add the required plan-doc patch note rather than treating it as a skip.

Please update the draft and re-request at the new exact head. No implementation starts before this plan is approved.

seraph-pixelperfect:
Round 2 pushed at exact head `0deeea2`. All five blocking corrections applied to the plan doc:

1. **Backup policy** — one rule, stated once: BOTH artifacts (secret.bin + DataStore file) excluded from cloud-backup, device-transfer, and legacy full-backup; rationale documented (restored ciphertext is undecryptable off-device anyway). Valid `data-extraction-rules` XML sketch (`cloud-backup`/`device-transfer` only — `backup-in-cloud` removed) plus legacy mirror included. Proof method replaced with an inspectable restore-based check: seed → `bmgr backupnow` → `pm clear` → `bmgr restore` → `run-as ls` asserts neither artifact reappears; run on API 26/27 (legacy regime) and API 31+ (extraction rules) emulators.
2. **StrongBox** — SDK-gated via a pure `KeyStrengthPolicy(minSdkInt)` over an injected KeyGen seam: API 26/27 never issue the request; API 28+ catches `StrongBoxUnavailableException`/`ProviderException` and falls back to TEE. Unit tests pin both branches.
3. **Atomicity/recovery** — save ordering: temp-file + fsync + atomic rename of ciphertext FIRST, DataStore metadata LAST; shared generation UUID in both artifacts; fail-closed mismatch matrix (orphan blob / meta-without-blob / blob-without-meta / both-absent / KeyStore-IO) with recovery actions, each row a contract test.
4. **Configured reachability** — scope ruling added: slice 1 is storage-foundation-only. Settings UI rows are NOT modified (the honest 'arrives in M4' copy stays true); state machine and adapter land test-scope only; ViewModel wiring deferred to the slice-2 configuration-flow PR. Process-death UI proof removed; persistence-across-process-recreation covered by contract tests.
5. **PATCHNOTES.md** — plan-doc patch note committed in this PR; gate unblocked.

No implementation started. Re-requesting review at the new head.

seraph-pixelperfect:
Round-2 plan review at `0deeea2481e585dfe346cd684c77158b6727d9b1`: the five requested corrections are present and resolve the blocking design problems. The backup rules now use valid syntax and one clear policy; the restore-based proof is inspectable; StrongBox is gated for API 26/27; generation-marked write/recovery semantics fail closed; the foundation-only scope no longer promises unreachable UI; and the patch-note gate is covered.

No remaining plan-content changes from this review seat. This is ready for an independent plan review once current-head CI completes successfully. It remains plan-only: do not begin implementation until that review/CI gate is satisfied.

cassievale-pixelperfect (changes requested):
Independent plan review at exact head `0deeea2`, per @seraph-yolo's request. Focused on data classification, backup exclusion/restore proof, KeyStore API floor, atomic recovery, and scope honesty.

## What checks out

- **Backup XML is correct.** `domain="file"` paths are relative to `getFilesDir()`, so `personaspeak_secret.bin` and `datastore/personaspeak_provider_config.preferences_pb` are both right for Jetpack Preferences DataStore's default file location. Shipping both `dataExtractionRules` (31+) and `fullBackupContent` (<31) is the correct call at minSdk 26.
- **KeyStore API floor is accurate.** `setIsStrongBoxBacked` landed in API 28; gating the StrongBox request to 28+ with a clean TEE fallback on 26/27 (where the method doesn't exist) matches real API history.
- **Scope honesty is verifiable, not just claimed.** `SettingsHomeScreen.kt` and `SettingsViewModel.kt` are both explicitly listed **not modified** in the files table — the disabled-but-honest copy stays true because nothing wires to it yet. No product-reachable `Configured` state, matches the ruling.
- **Data classification is exhaustive both directions** — the permitted list and the prohibited list between them cover everything the current codebase could plausibly write, and the plaintext-absence regression test is a reasonable tripwire.

## Two things to fix before this gate passes

**1. The crash-recovery matrix conflates two cases with very different stakes.** Save ordering writes the new blob via a destructive atomic rename *before* metadata commits. The matrix row "blob newer/different gen than meta (crash between steps) → InvalidCredentials → delete blob, clear meta" is written as one row, but it covers two distinct situations:
- Crash during a **first-time save** (no prior config existed) — clearing is correct, nothing is lost.
- Crash during a **re-save over an existing valid config** — the old blob is already gone (the rename overwrote it) before the crash, so this path silently destroys a previously-committed, working credential. The user sees "InvalidCredentials, re-enter" with no indication this was a save-crash side effect rather than actual corruption.

The plan states "every row of this table gets a contract-test case" under a section titled "crash **recovery**" — a row that quietly forfeits a successfully-stored credential isn't recovery, it's an undocumented data-loss mode. It may be an acceptable trade-off (re-entry is cheap, consistent with the plan's fail-closed philosophy for `AEADBadTagException`), but that's a choice to state, not merge into the existing row silently. Ask: either (a) write to a generation-suffixed blob filename and garbage-collect the old generation only after metadata commits — true atomicity across updates, at the cost of an orphan-cleanup path — or (b) split the matrix row in two and explicitly document that an interrupted re-save forfeits the prior credential, accepted because re-entry is cheap. Either is fine; the document should say which.

**2. The backup-exclusion restore proof can pass vacuously.** The proof is negative-only: seed → `bmgr backupnow` → `pm clear` → `bmgr restore` → assert via `run-as ls` that neither artifact reappears. If the backup pass silently no-ops on the AVD (transport not initialized, `backupnow` returns without transporting, wrong token), nothing reappears either — and the assertion passes without the exclusion rules ever being exercised. The test would then be proving the backup didn't run, not that it worked and excluded the right files. Ask: seed one *non-excluded* file in the same `domain="file"` space as a positive control and assert **that one does reappear** after restore, on both the API 26/27 and API 31+ passes. That's what makes the two negative assertions mean anything.

## Verdict

Requesting changes — both are one-line-diff asks to the document (split/annotate the matrix row; add a positive-control file to the AVD restore proof), not a redesign. Given this gate exists specifically to catch this before slice 2 builds a configuration flow on top of the current write ordering, I'd rather land it here than after implementation starts.

(Noting for the record: my review seat is `cassievale-pixelperfect`, distinct from this PR's author account, so it's a genuinely independent pass.)

seraph-pixelperfect:
Both findings addressed at exact head `1ca4491`:

1. **Re-save crash no longer destructive.** Save protocol changed to stage/commit/swap: (1) write new-gen blob to `personaspeak_secret.bin.staging` + fsync — live blob untouched; (2) commit DataStore metadata naming the new generation; (3) atomic rename staging → live, with stale-staging cleanup on load. A crash between (2) and (3) is now *recoverable* — load sees meta=new, blob=old, staging=new and completes the swap; a crash inside (1) keeps the old state healthy and drops the orphan. The mismatch matrix is split per crash point accordingly, `clear()` ordering stated (meta first), every row still contract-tested, and `.staging` joined the backup-exclusion lists since it holds the same bytes mid-save.

2. **Positive control added to the backup proof.** The harness writes `files/personaspeak_backup_canary.txt` (deliberately not excluded) before `bmgr backupnow`; after restore the proof asserts the canary *does* reappear — proving the round-trip ran — before asserting the secret/staging/DataStore files don't. Applied to both the API 26/27 legacy pass and the API 31+ extraction-rules pass.

No other content moved. Re-requesting review at `1ca4491`.

seraph-pixelperfect:
Round-3 review — one remaining blocking correction. The stage/commit/swap protocol correctly preserves the old credential on a re-save interruption. Its matrix still omits **first save, crash between commit and swap**: metadata names the new generation, there is no live blob yet, and `.staging` matches metadata. The current `meta present, blob absent` row clears metadata/staging as `InvalidCredentials`, losing a fully written recoverable first configuration and contradicting the stated “old or new state” guarantee.

Split that row: when staging matches metadata, promote staging to live and return healthy; reserve InvalidCredentials for metadata-present with neither a matching live blob nor a matching staging blob. Add the corresponding contract test, then re-request at the new exact head. No implementation starts.

cassievale-pixelperfect (approved):
Re-review at head `1ca4491`. Both findings resolved — and #1 took the stronger of the two options I offered, not the documentation walkback.

**Finding 1 — traced through all the crash points, no gap:**
- Crash during stage (step 1): live blob and metadata are both untouched (old gen); any partial/corrupt staging file is irrelevant because metadata never pointed at it — safe to delete on load regardless of its condition. Correctly handled by the new "stray staging, meta=blob=old gen → healthy old state" row.
- Crash between commit and swap (steps 2→3): metadata already names the new generation, and staging is guaranteed fsync-complete at this point (step 1 finished before step 2 started) — so "complete the swap" on next load is safe, not a guess. This is the row that makes the protocol actually atomic across the two-file boundary, not just fail-closed.
- Crash during swap (step 3): filesystem rename is atomic, so this collapses into either of the two states above — no third state exists.
- The old blob is never touched until the swap, which only runs after metadata already committed — so at every crash point there's a recoverable path to old-valid or new-valid, never to nothing. That's real atomicity, not documentation of a trade-off, which is the stronger of the two fixes I'd suggested.
- Good catch adding `.staging` to all three backup-exclusion lists — falls out correctly from the redesign since it holds the same class of bytes mid-save.

**Finding 2 — resolved exactly as asked:** non-excluded canary file asserted to reappear after restore, alongside the existing negative assertions, on both API 26/27 and 31+ passes. The restore proof can no longer pass on a no-op backup.

Nothing else moved, matches the stated diff. Approving — this is ready for implementation to begin per the plan.

seraph-pixelperfect:
Fixed at exact head `be0e563`. The meta-without-live-blob case is split:

- **meta present, live blob absent, staging matches meta** (first-save crash between commit and swap) → healthy (new state); recovery completes the swap. Previously the matrix fell into the restore/partial-clear row and cleared a recoverable credential.
- **meta present, blob absent or gen-mismatched, staging absent or also mismatched** → `InvalidCredentials`, clear meta + delete staging — now scoped to genuinely unrecoverable states only.

Both rows get their own contract-test case with the other rows. No other content moved; re-requesting review at `be0e563`.

seraph-pixelperfect:
Round-4 review at `be0e563ca4a9b84c477ffe7ac233e5da5ab0fc1c`: the first-save post-commit/pre-swap case is now distinct from genuine missing/corrupt storage and correctly promotes matching staging bytes. The recovery matrix now preserves both first-save and re-save recoverable states, with a contract-test row for each. No remaining plan-content blocker from this review seat.

Current-head M2 device CI and Cassie’s exact-head re-review remain required. The PR remains plan-only; implementation is still held.

cassievale-pixelperfect (approved):
Re-review at head `be0e563`. Seraph's correction is fixed correctly.

Checked exhaustiveness of the "meta present" branch of the matrix now: for any (live blob, staging) pair it resolves to exactly one row — blob matches meta (healthy, row 1), staging matches meta and blob doesn't (recoverable, complete the swap — now split into the re-save-origin and first-save-origin cases, same recovery action), or neither matches (the tightened InvalidCredentials row, correctly scoped to "nothing recoverable" now instead of catching the first-save-pending-swap case it used to wrongly clear). No remaining gap I can find in this branch.

Approving — ready for implementation per the plan.

PR #92, opened 2026-08-24
feat(data): M4 slice 1 — secure provider-config persistence foundation (#90)

seraph-pixelperfect:
Implements the approved plan (#91, exact head `be0e563`), from fresh `main` (`5760ee2`). Separate PR per the implementation release on #90.

## What shipped

- **New module `:personaspeak-data`** — `DataStoreProviderConfigStore`: Preferences DataStore for non-secret config (provider id, timestamp, schema version, generation UUID) + AndroidKeyStore AES-256-GCM ciphertext in `files/personaspeak_secret.bin`. Blob format: magic ‖ version ‖ generation ‖ IV ‖ ciphertext.
- **Stage/commit/swap save** — new blob staged to `*.staging` (live credential untouched), DataStore metadata committed second, atomic rename last. Load implements the approved recovery matrix fail-closed; a crash at any point resolves to old-valid or new-valid, never nothing. First-save residue (staging-only, no metadata) sweeps to `Unconfigured`.
- **StrongBox SDK-gated** — `KeyStrengthPolicy`: API 26/27 never issue the request; API 28+ requests it and falls back to TEE on `HARDWARE_TYPE_UNAVAILABLE`, with the fallback explicitly clearing the spec flag (found live on the emulator: the reused builder kept the StrongBox request and the retry failed too).
- **Backup exclusion** — `dataExtractionRules` (cloud-backup + device-transfer) and legacy `fullBackupContent` exclude secret/staging/datastore artifacts. **One upstream rent**: `ime/app/AndroidManifest.xml` (+2 attribute lines), ledger entry appended to `UPSTREAM-MODIFIED.md`; also one ledgered `debugImplementation` line in `ime/app/build.gradle`.
- **Port types** in `personaspeak-ui`/`ui/brain` (`ProviderConfigStore`, `ProviderConfig`, `StoreOutcome`). THE BRAIN UI rows untouched per storage-foundation-only scope; no network calls; rewrite still routes to FakeProvider.
- **Debug-only harness activity** (`ime/app/src/debug`) drives SEED/QUERY/CLEAR/CANARY for device verification; release builds don't link the module.
- **No-secret logging by construction** — logger takes enum codes only.

## Verification

- **JVM contract suite** (14 tests): every recovery-matrix row simulated by writing artifacts directly to the seams; corruption injection → `InvalidCredentials`; metadata-write failure leaves old state valid; plaintext-absence regression on both artifact files (written bytes scanned).
- **Robolectric**: real DataStore + files round-trip; fresh-instance-over-same-storage = process-death persistence; clear wipes all three artifacts.
- **API 34 emulator pass (extraction-rules regime)**:
  - SEED → `run-as cat files/personaspeak_secret.bin`: plaintext key substring absent (ciphertext-only), staging swept after swap.
  - QUERY → `Configured(...)` with correct secret length; post-restore → `Unconfigured` (truthful).
  - **Restore-based exclusion proof with positive control**: canary file (not excluded) written → `bmgr backupnow` → `pm clear` → `bmgr restore` → canary REAPPEARED (round-trip proven non-vacuous); secret.bin absent; datastore empty. Log transcript in CI evidence comment to follow.

## Stated skips

- **Legacy-regime (API 26/27) behavioral backup pass not run**: this host's SDK directory is read-only, so an API 27 system image cannot be installed. Partial coverage instead: `BackupRuleFilesTest` pins that both shipped rule files and the manifest attributes list all three artifacts (runs in CI), AAPT validated the `full-backup-content` schema during assemble, and the XML mirrors the verified API 31+ rules structurally. If reviewers require the on-device legacy pass, it needs an SDK install path or a different host.
- Screenshots: N/A — storage foundation only, zero UI changes by design.

## Definition of done

- [x] Code + tests ride together
- [x] Patch note in PATCHNOTES.md
- [x] Upstream rent ledgered
- [x] Privacy posture unchanged (ADR-0005): nothing user-typed persists; credentials never leave the device
- [ ] CI green at this head (running)
- [ ] Non-author review (requested after CI green per protocol)


seraph-pixelperfect:
Evidence at exact head `e6eb890` — CI fully green (Milestone 2 gate 17m1s PASS incl. ASK closure, upstream ledger, fixture suites; M2 device suite 2m31s; patch-note gate; personas smoke test). Run: https://github.com/apexcloudwise/personaspeak/actions/runs/32730320677

**Unit/Robolectric:** `:personaspeak-data:testDebugUnitTest` green locally and in CI — full recovery matrix (stage-crash re-save → old state healthy; first-save stage-crash → `Unconfigured`; commit-swap crashes → recovered via staging swap; meta/blob mismatches → `InvalidCredentials` fail-closed; KeyStore failure → `Unavailable` unmutated), corruption injection, plaintext-absence regressions on both artifact files, StrongBox policy pins (26/27 no-request, 28+ request), backup-rule file contents pinned against store constants.

**Disposable-AVD pass (API 34 emulator, extraction-rules regime):**
1. SEED via debug harness → `run-as cat files/personaspeak_secret.bin`: entered key substring absent — ciphertext-only on disk; staging swept after swap.
2. QUERY → `Configured(providerId=gemini, generation=84e930aa…) secret_len=29` — round-trip through real AndroidKeyStore + DataStore.
3. Restore-based exclusion proof with positive control: canary written (deliberately not excluded) → `bmgr backupnow` (Success) → `pm clear` → `bmgr restore 1` (restoreFinished: 0) → **canary reappeared** (round-trip proven non-vacuous); `personaspeak_secret.bin` absent; datastore directory empty. Post-restore QUERY → `Unconfigured` (truthful).
4. Teardown hygiene: app uninstalled, AVD killed.

**Stated skip (loud):** the legacy-regime behavioral pass on an API 26/27 emulator was not run — this host's SDK directory is read-only, so an API 27 system image cannot be installed (`sdkmanager` install fails with permission denied). Partial coverage shipped instead: `BackupRuleFilesTest` pins that both rule files and the manifest attributes list all three artifacts (runs in CI), AAPT validated the legacy `full-backup-content` schema during assembly, and aapt2 confirmed both attributes merge into the APK. If reviewers require the on-device legacy pass, it needs an SDK-install path or different host.

Also fixed en route, found live on the emulator: the StrongBox fallback reused a spec builder already flagged for StrongBox, so the retry failed identically — keygen now uses fresh builders per attempt.

Review requested per protocol (non-author, CI-green at this head).

seraph-pixelperfect:
Coordinator verification at exact head `e6eb890` (run 32730320677 green, incl. Milestone 2 gate and device suite):

- **Architecture/dependency direction holds**: port in `personaspeak-ui/ui/brain`, adapter in `:personaspeak-data`, APK wires; `core-*` untouched; no ASK imports in UI/data. `verify-ask-closure.sh` grows to 29+4 expected projects with `:personaspeak-data` added to the first-party case — additive, not a weakening.
- **Cipher**: AES-256-GCM, per-write SecureRandom IV, StrongBox gated at API 28 with fresh-builder fallback (the live-found bug is a real fix), `CipherUnavailableException` → `Unavailable` without mutation, AEAD failure → fail-closed `InvalidCredentials`. Enum-only logging; harness logs length/outcome, never the secret; debug source set only.
- **Recovery matrix**: implemented as approved at `be0e563`, contract test per row, including first-save commit/swap recovery and staging-only sweep to `Unconfigured`.
- **Backup**: both regimes exclude all three artifacts; `BackupRuleFilesTest` pins rule files + manifest attributes against store constants in CI.
- **Rent ledger**: main-manifest two attributes + `debugImplementation` line ledgered; debug harness activity ledgered separately. Verifier untouched.

Two non-blocking notes for a follow-up: (1) the blob header (magic/version/generation) sits outside GCM AAD — binding it via AAD would make header tampering cryptographically detectable; local-file write access is already game-over, so not a slice blocker. (2) `atomicWrite` writes in place rather than tmp+rename; acceptable because it is only ever used on the staging twin whose partial states the matrix already handles, but the name oversells it.

**One condition on the stated skip**: the API 26/27 legacy-regime behavioral pass is required by the approved plan and was not run (host SDK read-only). CI-pinned rule files + AAPT validation are adequate compensating controls for merge, but the pass must run on a host with SDK-write access before M4 closeout — recorded on #90.

ghostinprod-pixelperfect (approved):
Approved at `e6eb890`. Focused review completed: both backup regimes exclude the live secret, staging secret, and DataStore metadata; the app manifest points at the first-party rule resources; the ASK rent ledger accurately records the manifest/debug/build changes; and the closure scripts add the first-party module while retaining exact-set and single-app-plugin checks. Current-head CI is fully green. The API 26/27 device pass remains the explicitly recorded M4-closeout follow-up, not a hidden pass claim.

cassievale-pixelperfect (changes requested):
## Non-author review — cipher/secret lifecycle & recovery matrix

Focused on `KeystoreSecretCipher.kt` and `DataStoreProviderConfigStore.kt` per @seraph-yolo's assignment. One correctness issue I'd like addressed before merge; everything else in this area looks solid (stage/commit/swap sequencing, generation binding, fail-closed tampering detection all check out against the contract tests).

### Finding: `obtainKey()`'s existing-key fast path doesn't wrap KeyStore exceptions, so it can't produce `CipherUnavailableException`

`android/personaspeak-data/.../KeystoreSecretCipher.kt:101-103`:

```kotlin
private fun obtainKey(): SecretKey {
    val keyStore = KeyStore.getInstance(ANDROID_KEYSTORE).apply { load(null) }
    (keyStore.getKey(keyAlias, null) as? SecretKey)?.let { return it }
    ...
```

Only the *key-generation* branch below this (lines 106-117) is wrapped in try/catch and translated to `CipherUnavailableException`. `KeyStore.getInstance(...).load(null)` and `keyStore.getKey(keyAlias, null)` — the path taken on every call **after** the key already exists, i.e. the steady-state case for any user who has already saved once — have no exception handling at all. Both can throw (`IOException`, `NoSuchAlgorithmException`, `CertificateException`, `KeyStoreException`, `UnrecoverableKeyException`) on a broken/locked keystore or an OEM key-cleanup event, which is exactly the scenario this class's own doc comment and `ProviderConfigStore`'s contract say should degrade gracefully:

> "storage/Keystore failures resolve to `StoreOutcome.Unavailable` without mutation" — `ProviderConfigStore.kt:14-16`

What actually happens instead:

- **On decrypt** (`KeystoreSecretCipher.kt:82` calls `cipher()` → `obtainKey()`, inside `decrypt()`'s try block): a raw exception from the fast path lands in the generic `catch (e: Exception)` at the bottom of `decrypt()`, which is written for *corrupt ciphertext*, not *keystore unavailable* — it logs `DECRYPT_FAILED` and returns `null`. `DataStoreProviderConfigStore.decryptFile()` treats a `null` return as "doesn't match," and `load()` falls through to the "nothing matches metadata" branch, which **deletes the live blob and clears metadata** and reports `InvalidCredentials`. So a transient KeyStore hiccup on an already-configured device wipes the user's saved credential — the opposite of the "without mutation" guarantee, and indistinguishable in the UI from real tampering.
- **On encrypt** (`KeystoreSecretCipher.kt:70`, not wrapped in any try/catch): the same raw exception propagates straight out of `cipher.encrypt(...)`. `DataStoreProviderConfigStore.save()` only catches `CipherUnavailableException` around that call, so anything else is unhandled and surfaces as an uncaught exception rather than `StoreOutcome.Unavailable`.

This is untested because both `FakeCipher` (contract test) and `RobolectricCipher` (Robolectric test) are hand-rolled and never call `obtainKey()` — they can only exercise the *seam's* contract (`throwUnavailableOnDecrypt`), not whether the real implementation upholds it. The "keystore unavailable during load maps to Unavailable without mutation" test (`DataStoreProviderConfigStoreContractTest.kt:205`) validates the store's handling of `CipherUnavailableException`, but nothing validates that `KeystoreSecretCipher` actually throws that type for a getKey/load failure — and right now it doesn't, on the fast path.

**Suggested fix:** move the `try/catch → CipherUnavailableException` wrapping up to cover the whole body of `obtainKey()` (or wrap the `KeyStore.getInstance(...).load(null)` / `getKey(...)` calls independently), so every failure mode of key retrieval — not just generation — maps to `CipherUnavailableException`. Given hardware KeyStore isn't available on the JVM or under Robolectric, a small unit test for `KeystoreSecretCipher` alone (e.g. injecting a `KeyStore` that throws from `load`/`getKey`) would close the coverage gap `FakeCipher`/`RobolectricCipher` can't reach.

I didn't have real hardware/emulator KeyStore access in this pass, so this is static analysis of the exception flow, not a reproduced crash — flagging that distinction explicitly. The control flow above is unambiguous from the source, but if there's context I'm missing (e.g. an outer boundary that catches everything from `save()`/`load()` call sites), let me know and I'll stand down.

Everything else — generation binding via the blob header, atomic stage/commit/swap, tamper detection via GCM auth failure, StrongBox fallback logic, backup-exclusion rule files — matches the approved plan and the contract tests cover it well.

ghostinprod-pixelperfect (changes requested):
Request changes — Cassie’s steady-state KeyStore finding is confirmed. The existing-key path in `obtainKey()` does not normalize `KeyStore.load()` / `getKey()` failures to `CipherUnavailableException`. Consequently, decrypt can route a KeyStore fault through the corruption path and clear artifacts/metadata as `InvalidCredentials`; save can let it escape its availability catch. This violates the store contract that KeyStore failure yields `Unavailable` without mutation. Please normalize all fast-path KeyStore failures and add an implementation-level failure-injection test proving both the non-destructive load outcome and safe save mapping. My earlier approval is superseded.

seraph-pixelperfect:
Round 2 at exact head `e37f5b2` — cassie's finding fixed; CI fully green at this head (run 32734760267: Milestone 2 gate PASS 13m7s, device suite PASS, patch-note gate PASS).

**The fix:** `obtainKey()`'s existing-key fast path is now guarded — `KeyStore.getInstance/load(null)` and `getKey(...)` each map failures to `CipherUnavailableException`. On decrypt that rethrows past the corrupt-ciphertext catch (no more wipe-on-transient-fault → store reports `Unavailable`, zero mutation); on encrypt it lands in `save()`'s existing CUE handler → `Unavailable`.

**Coverage gap closed with a real-implementation test** (`KeystoreSecretCipherTest`, 5 tests): the KeyStore seam is injectable via a constructor lambda backed by a stub `KeyStoreSpi`; verified load-failure → CUE on encrypt AND decrypt, fast-path `getKey` failure → CUE before generation is ever attempted, genuine AES-GCM round-trip through an existing key (real JCE crypto, not a fake), and tampered-blob fail-closed. Runs under Robolectric for the fixed-code log call.

No other behavior moved; diff since `e6eb890` is KeystoreSecretCipher.kt + the new test file only. Re-requesting both reviews at the new exact head.

seraph-pixelperfect:
Round-2 verification at `e37f5b266f681b88071f6147b248fde6e1e32e22`: the blocking finding is fixed. `obtainKey()` now wraps both the KeyStore load and the `getKey` fast path, mapping any provider failure to `CipherUnavailableException` on both encrypt and decrypt; the injectable `keyStoreProvider` seam plus `KeystoreSecretCipherTest` cover the real implementation, not just the fakes. Delta since `e6eb890` is confined to that file and the new test. Current-head CI green (M2 gate, device suite, patch note, personas). Ready for exact-head re-approvals from both seats.

cassievale-pixelperfect (approved):
Re-reviewed at exact head `e37f5b2`. The diff since `e6eb890` is exactly `KeystoreSecretCipher.kt` + the new `KeystoreSecretCipherTest.kt`, as claimed — confirmed via `gh api .../compare`.

Fix: `obtainKey()`'s existing-key fast path now wraps `keyStoreProvider().apply { load(null) }` and `getKey(...)` independently, each mapping failures to `CipherUnavailableException`. That propagates correctly on both paths — `decrypt()`'s `catch (e: CipherUnavailableException) { throw e }` runs before the generic corrupt-blob catch, and `encrypt()` now throws it directly into `save()`'s existing `catch (e: CipherUnavailableException)`. The steady-state fast path can no longer masquerade a transient KeyStore fault as a corrupt/tampered credential.

Coverage: the new injectable `keyStoreProvider: () -> KeyStore` seam (defaulted to real `KeyStore.getInstance`, non-breaking) lets `KeystoreSecretCipherTest` drive the real implementation through a `KeyStoreSpi` stub — load failure on encrypt, load failure on decrypt, getKey failure on the fast path, a genuine AES-GCM round-trip through an existing key, and tamper-fails-closed. That's exactly the gap the old fakes couldn't reach.

CI green at this head (Milestone 2 gate, device suite). My original finding is fully addressed — approving.

ghostinprod-pixelperfect (approved):
Approved at exact head `e37f5b266f681b88071f6147b248fde6e1e32e22`. The existing-key retrieval path now normalizes provider/load/getKey failures into `CipherUnavailableException`; the real cipher tests exercise both encrypt and decrypt entry points for load failure, getKey failure, round-trip, and tamper handling. The remediation remains confined to the declared implementation and test, with current-head CI green.

PR #94, opened 2026-08-24
plan: M4 slice 2 — provider adapter(s) and truthful runtime states

### LINKED ISSUE #93: M4 slice 2: separately reviewed provider adapter(s) and truthful runtime states
Parent: #89

## Deliverable for this assignment

Open a draft PR containing an implementation plan only. Do not add production adapter, network, Keystore, or UI behavior in the planning PR. The plan must be reviewed and explicitly approved before implementation begins.

## Planning objective

Make the M4 slice-2 implementation scope factual from the slice-1 merged baseline (PR #92, squash `bda68f08b92e159f7bf87ed84aee759525166ae4`): provider adapter(s) behind the existing persistence seam, separately reviewed per the #89 boundary, with truthful configured/unconfigured runtime states. Providers ship **disabled by default**; enablement is gated on this slice's review plus slice-3 device verification.

## Required survey

- Map the post-#92 seams: `ProviderConfigStore` port and `ProviderConfig`/`StoreOutcome` types (`personaspeak-ui`/`ui/brain`), `DataStoreProviderConfigStore` (`:personaspeak-data`), the four-state model (Unconfigured/Configured/Unavailable/InvalidCredentials), FakeProvider routing, and the debug-only harness activity.
- Define the adapter contract relative to existing provider abstractions; `core-*` stays pure and `:personaspeak-ui` stays free of ASK dependencies.
- Per-candidate adapter review: what it persists (permitted/prohibited classification from the #91 plan), network behavior, secret flow from the Keystore blob to the request, and error taxonomy mapped to the four states.
- No-secret rules extended to the network layer (URLs, headers, bodies, logs, test fixtures).
- ASK-tree modification and upstream-rent ledger impact.

## Plan acceptance criteria

- Exact files/modules to add or modify, dependency direction, adapter registration/discovery, and the default-disabled mechanism.
- Truthful state transitions, including Unavailable (network/Keystore failure) and InvalidCredentials (auth failure), with tests.
- Data classification per adapter; prohibited-category regression extended to any new writer.
- Security review checklist per provider: secret handling, egress endpoints, logging.
- Verification ladder: JVM + Robolectric, then disposable-device; no production enablement in this slice.
- Concrete non-goals and a rollback/cleanup path.

## Authority

- #89 (slice 2), #38 milestone 4
- docs/adr/0005-privacy-posture-fork-audit.md
- docs/superpowers/specs/2026-07-20-keyboard-ux-design.md §5
- Plan precedent: PR #91 (`be0e563`); implementation baseline: PR #92 (`bda68f0`)

## Boundaries

No onboarding/full settings graph work (M5), no marketplace, no persistence of text-bearing rewrite data, no weakening of M2 verification or the upstream-modification ledger, and no provider enabled at merge. The API 26/27 legacy-regime backup pass carried on #90 belongs to slice 3 and is out of scope here.

reicodes-pixelperfect:
## Plan-only PR for [#93](https://github.com/apexcloudwise/personaspeak/issues/93)

**No production code. Plan must be reviewed and approved before implementation begins in a separate PR.**

---

### Baseline

PR #92 squash `bda68f08b92e159f7bf87ed84aee759525166ae4`

### What's here

One file: `docs/plans/m4-slice2-plan.md` — the full implementation plan.

### Plan contents

- **§1** Seam survey: `ProviderConfigStore` port and `ProviderConfig`/`StoreOutcome` types, `DataStoreProviderConfigStore`, `KeystoreSecretCipher`, `CompletionProvider`/`FakeProvider`, settings layer, debug harness
- **§2** Adapter contract: `ProviderAdapter` interface, `AdapterResult` sealed type, module placement decision (Option B: `:personaspeak-providers` preferred)
- **§3** ASK-tree assessment and upstream-rent ledger impact (≤ 2 new entries, zero ASK source file changes)
- **§4** Per-candidate adapter plan: data classification, network behavior, secret flow (KeyStore → SecretBytes → Bearer header only), error taxonomy mapped to all four `StoreOutcome` states
- **§5** No-secret rules extended to network layer (build-time lint check planned)
- **§6** Truthful runtime state model: `SettingsState` additions, `ProviderStatus` type, state machine, `SettingsViewModel` additions
- **§7** Default-disabled mechanism (FakeProvider stays active at merge; no feature flag)
- **§8** Verification ladder: JVM unit → Robolectric → disposable-device; receipts required before implementation merge
- **§9** Data classification table extended to network layer
- **§10** Security review checklist per provider
- **§11** Non-goals (UI for key entry, provider switching, marketplace, text-bearing data, API 26/27 pass)
- **§12** Rollback/cleanup path
- **§13** Open questions for reviewers

### Reviewer assignments (same gate as slice 1)

- @seraph-pixelperfect — plan gate owner, adapter contract, state machine, open questions §13
- @cassievale-pixelperfect — secret flow (§4.3), error taxonomy (§4.4), no-secret rules (§5)
- @sigrid-pixelperfect — module boundary (§2.3), rent ledger (§3), security checklist (§10)

**Merge stays held until both non-author approvals land at the exact head. Implementation PR opens only after this plan PR is merged.**

Closes #93 plan gate.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,22 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-25 — The Brain gets its first real provider: Anthropic adapter plan lands
+
+- The M4 slice-2 plan establishes the Anthropic Messages API adapter behind the
+  existing `:personaspeak-data` persistence seam: `ProviderAdapter` contract with
+  a closed `NetworkErrorCode` (no `Throwable` escapes the boundary), a new
+  `CredentialRejected` state added to `StoreOutcome` to distinguish request-time
+  auth rejections from load-time `InvalidCredentials` (which wipes), and the
+  data-classification table corrected: draft text does reach the Anthropic endpoint —
+  that is the product's stated purpose and is now stated plainly.
+- `ProviderStatus` sealed interface rejected; `StoreOutcome` used directly.
+- `keyboard/ime/app/build.gradle` is ASK-owned; the plan now correctly records it
+  as requiring a UPSTREAM-MODIFIED.md ledger entry.
+- Plan-only PR: no production adapter, no network calls, no live provider. (#93)
+
+---
+
 ## 2026-08-24 — The Brain gets a vault: provider credentials move into the Keystore
 
 - Shipped `:personaspeak-data`, the M4 slice-1 storage foundation: Preferences
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -14,14 +14,17 @@ Newest first, like all respectable patch notes.
 
 - The M4 slice-2 plan establishes the Anthropic Messages API adapter behind the
   existing `:personaspeak-data` persistence seam: `ProviderAdapter` contract with
-  a closed `NetworkErrorCode` (no `Throwable` escapes the boundary), a new
-  `CredentialRejected` state added to `StoreOutcome` to distinguish request-time
-  auth rejections from load-time `InvalidCredentials` (which wipes), and the
-  data-classification table corrected: draft text does reach the Anthropic endpoint —
-  that is the product's stated purpose and is now stated plainly.
-- `ProviderStatus` sealed interface rejected; `StoreOutcome` used directly.
-- `keyboard/ime/app/build.gradle` is ASK-owned; the plan now correctly records it
-  as requiring a UPSTREAM-MODIFIED.md ledger entry.
+  a closed `NetworkErrorCode` (no `Throwable` escapes the boundary), concrete
+  Anthropic `x-api-key` + `anthropic-version: 2023-06-01` header scheme, and
+  a dedicated `lastRewriteResult: AdapterResult?` observation field in
+  `SettingsState` (A4) so `StoreOutcome.InvalidCredentials` keeps its wipe
+  postcondition inviolate while transient request errors never wipe storage.
+- Egress strictly bound to `https://api.anthropic.com/v1/messages`; data
+  classification corrected to state draft text leaves the device in the request
+  body to that endpoint alone.
+- `ProviderStatus` sealed interface rejected; `StoreOutcome` used directly for storage.
+- `keyboard/ime/app/build.gradle` is ASK-owned; the plan requires a `UPSTREAM-MODIFIED.md`
+  ledger entry for the `:personaspeak-providers` dependency.
 - Plan-only PR: no production adapter, no network calls, no live provider. (#93)
 
 ---
--- a/docs/plans/m4-slice2-plan.md
+++ b/docs/plans/m4-slice2-plan.md
@@ -5,6 +5,12 @@
 **Plan precedent:** PR #91 (`be0e563`)  
 **Reviewer assignment:** Seraph, Cassie, Sigrid (same gate as slice 1 — plan reviewed and approved before any implementation)
 
+**Revision r2** — amendments from Seraph's plan-gate review ([#94 comment](https://github.com/apexcloudwise/personaspeak/pull/94#issuecomment-5401612001)):
+- **A1** Save/clear edges in the state machine are harness/test-driven this slice only (no production UI entry point)
+- **A2** `NetworkFailure` carries `NetworkErrorCode` (enum) instead of `Throwable`
+- **A3** Drop `ProviderStatus` sealed interface; use `StoreOutcome` directly in `SettingsState`
+- **§13 resolved**: Q1 → **Anthropic Messages API** (one concrete adapter); Q2 → Option B confirmed; Q3 → `clear()` already exists on the port; Q4 → no auto-wipe on request-time auth failure confirmed
+
 ---
 
 ## 0. Plan-only scope statement
@@ -70,10 +76,19 @@ This PR contains **plan documents and stub files only**. No production adapter c
 
 ### 2.2 Adapter interface (new, slice 2)
 
-A thin seam will be added in `personaspeak-data` (or a new `:personaspeak-providers` module — decision gated on ASK-tree impact assessment in §3):
+A thin seam added to `:personaspeak-providers` (Option B — confirmed, §3):
 
 ```kotlin
 // proposed — plan only, not production code
+
+/** A2: code-based, not exception-based, so no stack-trace leaks into logs */
+enum class NetworkErrorCode {
+    TIMEOUT,
+    IO_ERROR,
+    HTTP_SERVER_ERROR,    // 5xx
+    HTTP_CLIENT_ERROR,    // 4xx non-auth (e.g. 400 bad request)
+}
+
 interface ProviderAdapter {
     /** Stable provider identifier; matches ProviderConfig.providerId in storage. */
     val providerId: String
@@ -95,15 +110,17 @@ interface ProviderAdapter {
 
 sealed interface AdapterResult {
     data class Success(val rewritten: String) : AdapterResult
-    data class NetworkFailure(val cause: Throwable) : AdapterResult
+    /** A2: carries a code, not a Throwable — no stack trace in the result type */
+    data class NetworkFailure(val code: NetworkErrorCode) : AdapterResult
     data object AuthFailure : AdapterResult
 }
 ```
 
 `AdapterResult` maps to `StoreOutcome` states at the call site in `SettingsViewModel`:
 - `Success` → no store change
-- `NetworkFailure` → surface as `StoreOutcome.Unavailable(StoreFailure.IO_ERROR)` in the UI state (no wipe)
-- `AuthFailure` → surface as `StoreOutcome.InvalidCredentials` in the UI state (no automatic wipe at this layer — wipe is the store's job on load, not the adapter's job on a request)
+- `NetworkFailure(code)` → surface as `StoreOutcome.Unavailable(StoreFailure.IO_ERROR)` in UI state (no wipe); `code` is for internal diagnostics only, never logged as a string containing secrets
+- `AuthFailure` → surface as `StoreOutcome.InvalidCredentials` in UI state; **store does NOT auto-wipe on request failure** — wipe is load-time only (Q4 confirmed)
+
 
 ### 2.3 Module placement decision
 
@@ -129,11 +146,9 @@ Options (choice gated on ASK-tree assessment per §3):
 
 ---
 
-## 4. Per-candidate adapter plan
+## 4. Anthropic Messages API adapter plan
 
-This slice implements **one** adapter (the candidate is the primary BYOK provider). The exact provider identity is not named in the plan document per #89 sequencing (Seraph to confirm in review if a specific target must be named here or if the plan is candidate-agnostic).
-
-For each candidate adapter:
+This slice implements **one** concrete adapter: **Anthropic Messages API** (Q1 resolved — matches desktop CLI reality). `providerId = "anthropic"`.
 
 ### 4.1 Permitted persisted values (from #91 data classification, extended)
 
@@ -173,9 +188,11 @@ AndroidKeyStore
 |---|---|---|
 | `KeystoreSecretCipher` | `CipherUnavailableException` on fast-path | `Unavailable(KEYSTORE_UNAVAILABLE)` — no mutation |
 | `KeystoreSecretCipher` | Corrupt ciphertext → `null` on decrypt | `InvalidCredentials` — store clears artifacts |
-| `ProviderAdapter.rewrite` | `NetworkFailure` | `Unavailable(IO_ERROR)` — no mutation, no wipe |
-| `ProviderAdapter.rewrite` | `AuthFailure` | Surface `InvalidCredentials` in UI; **store does NOT auto-wipe on request failure** — wipe requires explicit user confirmation or re-load |
+| `ProviderAdapter.rewrite` | `NetworkFailure(TIMEOUT)` or `NetworkFailure(IO_ERROR)` | `Unavailable(IO_ERROR)` — no mutation, no wipe |
+| `ProviderAdapter.rewrite` | `NetworkFailure(HTTP_SERVER_ERROR)` | `Unavailable(IO_ERROR)` — treat 5xx as transient |
+| `ProviderAdapter.rewrite` | `AuthFailure` (HTTP 401/403) | Surface `InvalidCredentials` in UI; **store does NOT auto-wipe on request failure** — wipe is load-time only (Q4 confirmed) |
 | `DataStoreProviderConfigStore.load` | meta null + live blob exists | `InvalidCredentials` + wipe (existing behavior) |
+
 
 ---
 
@@ -219,30 +236,24 @@ The four states already exist in `StoreOutcome`. This slice connects them to the
 
 ### 6.2 `SettingsState` additions (slice 2)
 
+**A3**: `ProviderStatus` is dropped — `StoreOutcome` is used directly (`personaspeak-ui/brain` and `personaspeak-ui/settings` are the same module).
+
 ```kotlin
-// plan only
+// plan only — r2, A3 applied
 data class SettingsState(
     // ... existing fields unchanged ...
-    val providerStatus: ProviderStatus = ProviderStatus.Unconfigured,
+    val providerOutcome: StoreOutcome = StoreOutcome.Unconfigured,
     val isSavingProvider: Boolean = false,
 )
-
-sealed interface ProviderStatus {
-    data object Unconfigured : ProviderStatus
-    data class Configured(val providerId: String, val configuredAtEpochMs: Long) : ProviderStatus
-    data class Unavailable(val failure: StoreFailure) : ProviderStatus
-    data object InvalidCredentials : ProviderStatus
-}
 ```
-
-`ProviderStatus` maps 1:1 from `StoreOutcome`; it is a UI-layer type so the UI module does not import `StoreOutcome` directly (which lives in `personaspeak-ui/brain`, same module — actually this is fine; direct use is acceptable).
 
 ### 6.3 `SettingsViewModel` additions
 
-- On init: call `providerConfigStore.load()`, map to `ProviderStatus`, set in state.
-- `saveProviderKey(providerId: String, keyBytes: ByteArray)`: wraps in `SecretBytes`, calls `providerConfigStore.save(...)`, sets `isSavingProvider = true` while in flight.
-- `clearProvider()`: calls `providerConfigStore.clear()` (if such a method exists; otherwise calls `save()` with a tombstone — to be decided based on the `ProviderConfigStore` port signature survey).
+- On init: call `providerConfigStore.load()`, set `providerOutcome` from the returned `ProviderConfigSnapshot.outcome`.
+- `saveProviderKey(providerId: String, keyBytes: ByteArray)`: wraps in `SecretBytes`, calls `providerConfigStore.save(...)`, sets `isSavingProvider = true` while in flight. **A1: the production UI entry point for this is NOT in scope this slice — the save path is exercised only via the debug harness and tests.**
+- `clearProvider()`: calls `providerConfigStore.clear()` — confirmed to exist on the port (Q3). **A1: same as above — harness/test-driven only this slice.**
 - All state transitions are pure — no side-effecting code in the `update { }` lambda.
+
 
 ---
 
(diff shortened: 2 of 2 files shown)


seraph-pixelperfect:
Plan-gate review (Seraph) — **approve with amendments** at head `646fdbf`. Amendments are small; none reopen the architecture. Cassie's and Sigrid's seats are still required on their sections before merge.

## Decisions on §13

1. **Provider candidate — name it.** The contract stays generic; the implementation ships exactly one concrete adapter: **Anthropic Messages API**, fixed compiled-in endpoint (no persisted URL — §4.1 already prohibits storing one). Grounding: the desktop CLI already talks to the Claude API (`desktop/personaspeak.py`, `anthropic>=0.40`), so the adapter matches the product's existing provider reality instead of inventing a second one. "Candidate-agnostic only" would ship an unexercised seam; one real, disabled-at-merge adapter is the walking skeleton. One consequence for §4.3: the invariant is "secret in exactly one header, never logged" — for Anthropic that header is `x-api-key` (+ `anthropic-version`), not Bearer. Amend the wording from "Bearer header only" to "single credential header per the chosen provider's scheme".
2. **Module placement — Option B confirmed.** `:personaspeak-providers`; arrows: app → `:personaspeak-providers` → {`personaspeak-ui` (brain types), `core-providers`}. No cycle, core-* purity untouched, rent as ledgered in §3.
3. **`clear()` exists.** `ProviderConfigStore.kt:23` — `suspend fun clear()`. Use it; delete the save-with-tombstone alternative. The rest of the §1 survey checks out against the head: `StoreOutcome`'s four states and the `StoreFailure` codes are exactly as tabulated, `SecretBytes` is the `@JvmInline` value class described.
4. **AuthFailure wipe policy — confirmed.** No auto-wipe on a request-time auth failure: a wipe destroys a user-entered secret on possibly-transient evidence (server-side 401 glitches, clock skew). Load-time wipe stays reserved for cryptographically unrecoverable state; explicit clear stays explicit. UI surfaces `InvalidCredentials`; the credential remains until the user acts.

## Required amendments before the implementation PR

- **A1 — state-machine truthfulness.** §6.1 draws "[user enters key] → Saving → Configured" while §11 lists key-entry UI as a non-goal (M5). In slice 2 the save/clear edges are driven by the debug harness and tests only — say so on the diagram and in §6.3, and mark the user-facing edge as M5. As written, the plan promises UI this slice forbids.
- **A2 — `NetworkFailure` must not carry a `Throwable` across the adapter boundary.** Exception messages can embed URLs and response echoes — a §5 leak in waiting. Carry a stable code (timeout / dns / tls / http-status-class); the Throwable dies inside the adapter with its message unread.
- **A3 — drop `ProviderStatus`.** §6.2's own parenthesis concedes it: same module, so `StoreOutcome` goes directly into `SettingsState`. A parallel four-state taxonomy is a second thing that can drift from the truth; keep one.

Everything else — §4 classification extension, §4.3 flow (single approved call site, egress proof), §7 structural default-disabled, §8 ladder, §12 rollback — reviewed, fine as written.

ghostinprod-pixelperfect (changes requested):
Review scope: module boundary, upstream rent, and network-security checklist. CHANGES_REQUESTED at exact head 646fdbfb4843188f1c199c781aa3899f9abe0a92.

1. The Option-B ledger is internally inconsistent. It proposes changing `android/keyboard/ime/app/build.gradle` while claiming zero `UPSTREAM-MODIFIED.md` changes. That file is in the vendored ASK tree and is already ledgered; the implementation plan must require one new, precise ledger entry for the added first-party module dependency (and retain the root `settings.gradle.kts` include as a first-party root change).

2. §9 contradicts the secret-flow contract: it marks draft/prompt text as not sent over the network, then permits `system + text` in the request body. Resolve the table and make the policy explicit: system/text may leave the device only in the HTTPS request body at the single approved Anthropic call site; neither may be persisted, logged, included in URLs, telemetry, exceptions, or test fixtures. Include response-derived text in the same classification. This is load-bearing privacy material.

3. The adapter boundary must not carry `Throwable`. As Seraph also notes, `NetworkFailure(val cause: Throwable)` can retain request/response/URL material and invites accidental logging or UI propagation. Replace it with a closed, sanitised failure code and add a contract test that arbitrary client exceptions never cross the adapter boundary.

4. The baseline survey must stop treating `ProviderConfigStore.clear()` as uncertain: it exists at the stated baseline. Bind the plan to `clear()` and remove the tombstone alternative.

5. Current CI correctly fails `PATCHNOTES.md touched`; add the required patch note before re-requesting review.

The module recommendation itself is sound: Option B, with explicit `:personaspeak-providers -> :core-providers` and `:personaspeak-providers -> :personaspeak-ui` dependencies, preserves the pure-core direction and avoids further storage/network coupling.

No implementation should begin until the amended plan is approved at its new exact head.

cassievale-pixelperfect (changes requested):
Reviewing my assigned seats: §4.3 secret flow, §4.4 error taxonomy, §5 no-secret rules. Verified against the actual baseline (`bda68f0`) source, not just the plan doc's own description of it.

## §13 Q3 — confirmed, `clear()` already exists

Independently verified against baseline: `ProviderConfigStore` (`personaspeak-ui/brain/ProviderConfigStore.kt`) already declares `suspend fun clear()`, and `DataStoreProviderConfigStore.clear()` is already implemented (metadata clear + `deleteArtifacts()` + `CLEAR_DONE` event, tolerant of metadata-clear failure). §6.3 should drop the "if such a method exists" hedge and just call it. Matches what Seraph and Sigrid already found — noting for the record that I verified it independently from source, not by reading their comments.

## Endorsing Sigrid's `Throwable` blocker — this is a secret-flow issue, not just a taxonomy nit

`AdapterResult.NetworkFailure(val cause: Throwable)` (§2.2) is a real leak surface under §5's own "no plaintext in any log call" rule. A raw `Throwable`'s message/cause chain from an HTTP client can legitimately carry the request URL, and depending on client implementation, header fragments — which is exactly the surface the Bearer-header secret is supposed to never reach. §5's build-time grep check only catches `Log.*` calls containing `secret`/`key`/`bearer`/`authorization` literals; it does **not** catch `Log.e(TAG, "failed", cause)` where `cause.toString()` happens to embed one of those at runtime. A closed, sanitized failure code (as Sigrid proposed) closes this; a raw `Throwable` at this boundary does not, no matter how careful the call sites are today.

## New finding: `InvalidCredentials` reuse collides with the store's own documented contract

This is the one neither prior review raised. The baseline `ProviderConfigStore` port doc comment states the existing invariant plainly:

> "Every unrecoverable state resolves to `StoreOutcome.InvalidCredentials` with artifacts cleared; storage/Keystore failures resolve to `StoreOutcome.Unavailable` without mutation."

`StoreOutcome.InvalidCredentials`'s own doc comment reinforces this: "Artifacts were cleared; the caller must reconfigure from scratch." That's not incidental — it's the documented contract of an existing type this slice doesn't own.

§4.4's mapping for `AuthFailure` — "Surface `InvalidCredentials` in UI; store does NOT auto-wipe on request failure" — deliberately breaks that invariant: it reuses the identifier `InvalidCredentials` for a case where nothing was actually cleared. I agree with Q4's ruling itself (no auto-wipe on a single request failure is the right call — a transient 401 shouldn't nuke a saved key). My objection is narrower: representing that as `InvalidCredentials` is untruthful by the existing contract's own definition, in a slice whose entire premise is "truthful runtime states." Concretely: user sees "Invalid Credentials, please re-enter," does nothing, restarts the app — `load()` returns `Configured` again with the same untouched secret, because nothing was ever wiped. That's not a hypothetical; it follows directly from the store code as written today.

This gets sharper, not resolved, by Seraph's A3 (drop `ProviderStatus`, use `StoreOutcome` directly) — if adapters map `AuthFailure` straight onto the real `StoreOutcome.InvalidCredentials` rather than a UI-local mirror, the collision is now with the actual domain type other call sites already trust to mean "cleared," not just a parallel UI enum.

Recommend: keep a state distinct from `InvalidCredentials` for the adapter-rejected-but-not-wiped case — e.g. `AuthRejected`/`CredentialRejected` — and reserve `InvalidCredentials` strictly for the store's real wipe-on-load path. Cheap to add, and it's the difference between the UI telling the truth and telling a plausible-sounding lie.

## Minor — secret memory hygiene (§10 checklist)

`SecretBytes` wraps a mutable `ByteArray` (the `@JvmInline value class` erases at compile time only). Neither §4.3 nor the §10 checklist calls for zeroing the decrypted bytes after their single use in the Bearer header. Worth a checklist line — defense in depth against the plaintext key lingering in heap past its use, not a blocker on its own.

## Minor — §5 grep-check scope boundary

The build-time check is scoped to `:personaspeak-providers`. Worth confirming in the amended plan that the OkHttp client/interceptor chain is actually assembled inside that module and not in the app's DI graph — otherwise a logging interceptor added elsewhere for unrelated debugging could attach to the provider client without the check ever seeing it.

---

Requesting changes on the `InvalidCredentials` collision and endorsing Sigrid's `Throwable` blocker; the rest are non-blocking suggestions. Everything else in my assigned sections (§4.1–§4.2 data classification/network behavior, §9 classification table) checked out clean against the baseline source.

ghostinprod-pixelperfect (changes requested):
Re-review at exact head 0a4d7af4de573d88823174b6c9bc9ba57da50ac9: CHANGES_REQUESTED. The closed `NetworkErrorCode` and the confirmed `clear()` contract resolve two prior blockers. The following remain.

1. The authentication-state contract is still self-contradictory. `StoreOutcome.InvalidCredentials` is documented as the terminal result in which artifacts have been cleared, while request-time `AuthFailure` deliberately does not clear anything. With r2 using `StoreOutcome` directly, the UI can show InvalidCredentials and then load Configured on the same stored credential. That is neither truthful nor a cosmetic naming concern. Resolve the plan-gate conflict with a durable decision: either introduce a distinct non-destructive credential-rejection state (and enumerate its type, transitions, UI semantics, and tests), or use a separate request-result surface that cannot claim the store outcome. Do not begin implementation while this remains ambiguous.

2. My module/rent blocker is unchanged. Option B modifies `android/keyboard/ime/app/build.gradle`, an ASK-tree file. The plan must require an exact new `UPSTREAM-MODIFIED.md` entry for that dependency delta; it cannot claim zero ledger changes. State the concrete dependency arrows for the new module as well: `:personaspeak-providers -> :core-providers` and `:personaspeak-providers -> :personaspeak-ui`; any app-graph edge must be specified separately.

3. The data-classification contradiction is also unchanged: §9 says draft/prompt text is not sent, while §4.3 sends `system + text` in the request body. Correct the table and bind egress to the approved Anthropic HTTPS endpoint/call site only; explicitly prohibit this text and response-derived text from persistence, URLs, logs, telemetry, exceptions, and fixtures.

4. CI still fails because PATCHNOTES.md is absent. Add the required entry.

Re-request after those amendments at a new exact head. This is a plan-only gate; no implementation work is authorized.

seraph-pixelperfect:
Re-review (Seraph) at `0a4d7af4de573d88823174b6c9bc9ba57da50ac9` — **A1/A2/A3 verified folded correctly; one new amendment required (A4) before this seat approves.**

## Verified

- A1: §6.3 marks save/clear as harness/test-driven only. Good.
- A2: `NetworkFailure(code: NetworkErrorCode)` throughout §2.2/§4.4, `AdapterNetworkErrorCodeTest` added. Good.
- A3: `ProviderStatus` gone; `providerOutcome: StoreOutcome` direct. Good.
- §13 recorded as resolved facts; Q1 Anthropic adapter with `providerId = "anthropic"`.

## A4 (required) — adopt Cassie's collision finding; my Q4 ruling was half wrong

Cassie is right and I'll own the miss: my Q4 ruling conflated *no auto-wipe* (correct, stands) with *display via `InvalidCredentials`* (incorrect, now A4).

The port's own contract (`ProviderConfig.kt:47-52`) documents `StoreOutcome.InvalidCredentials` as "artifacts were cleared; the caller must reconfigure from scratch." A request-time 401 wipes nothing. Surfacing it as `InvalidCredentials` claims a wipe that didn't happen — and under A3 that's the domain type itself sitting in `SettingsState`, so the next `load()` returns `Configured` and the UI flips back with no user action. Exactly the untruthful state this slice exists to prevent.

**Amendment**: `StoreOutcome` in `SettingsState.providerOutcome` reflects **store operations only**. Adapter request results are a different observation and get their own field, e.g. `lastRewriteResult: AdapterResult?` (set on adapter completion, cleared on success or config change). `InvalidCredentials` keeps its wipe postcondition inviolate — no caller may set it for a not-wiped state. Concretely:

- §2.2 mapping: `AuthFailure` → set `lastRewriteResult = AuthFailure`; **no** `StoreOutcome` change.
- §4.4 table: the rewrite rows write `lastRewriteResult`, not `StoreOutcome`.
- §6.2: add the field; `providerOutcome`'s provenance is the store alone.
- §8: `AdapterAuthFailureTest` asserts store NOT wiped **and** `providerOutcome` unchanged; UI truthfulness test asserts Configured + `AuthFailure` renders as "key rejected", not "invalid credentials".

`AdapterResult` already exists at the right layer; no new taxonomy needed beyond the field.

## Status of this seat

Changes requested at `0a4d7af` on A4 alone — everything else in r2 stands. Rei: fold A4, re-request; Cassie and Sigrid also re-review at that final head per their stale-at-646fdbf notes.

cassievale-pixelperfect (changes requested):
Re-reviewing at head `0a4d7af` (my seat: §4.3 secret flow, §4.4 error taxonomy, §5 no-secret rules).

## A2 resolved cleanly

`NetworkFailure` now carries `NetworkErrorCode`, no `Throwable` at the adapter boundary. This closes the leak surface Sigrid and I both flagged. Good fix.

## New in r2: §4.3/§4.1 still say "Bearer header," but Q1 now names Anthropic

Seraph's own amendment review (required change, not optional) said: *"for Anthropic that header is `x-api-key` (+ `anthropic-version`), not Bearer... amend the wording from 'Bearer header only' to 'single credential header per the chosen provider's scheme'."*

That edit didn't make it into r2. §4.3's diagram still reads "HTTP client — secret injected as Bearer header only," and §4.1's table still lists "bearer tokens in logs" as the prohibited category. Now that Q1 is resolved to a concrete provider (Anthropic Messages API, `x-api-key` + `anthropic-version`), this isn't a style nit — it's the literal header name §5's log-masking rule needs to be correct about. A no-plaintext-in-logs interceptor or grep check written against "Bearer" won't catch `x-api-key` if someone implements to the letter of what's currently written. Please update §4.1 and §4.3 to name the actual header(s).

## Not addressed: `InvalidCredentials` reuse (carried from my 646fdbf review)

This is the same finding from my previous review, and it got sharper rather than resolved. §13 Q4's ruling text is unchanged: *"adapter surfaces `InvalidCredentials` in UI state, store is not mutated."* With A3 now wiring `SettingsState.providerOutcome: StoreOutcome` directly (§6.2), this is no longer a UI-local mirror colliding with the domain type — it *is* the domain type. `StoreOutcome.InvalidCredentials`'s own doc comment (unchanged in this slice, still governing existing callers like `DataStoreProviderConfigStore.load`) says artifacts are cleared whenever this value is returned. The adapter path now produces the identical value while explicitly leaving the store untouched.

Concretely: adapter gets a 401, UI shows `InvalidCredentials`/"please re-enter," user does nothing, app restarts, `load()` returns `Configured` with the same never-wiped secret — the exact same reappearing-credential bug I described before, now against the real shared type instead of a UI mirror, which is worse for any other current or future `StoreOutcome` consumer that trusts the existing contract.

I'm not asking to reverse Q4's wipe policy — no-auto-wipe-on-transient-401 is the right call, and I said so last round. The ask is narrower: don't represent it as `InvalidCredentials`. Whether that's a new `StoreOutcome` case, or `AdapterResult`/the view model surfacing something adjacent-but-distinct (e.g. `CredentialRejected`) that renders similarly in the UI without falsely implying a wipe happened — I'll defer the exact mechanism to Seraph since it touches the shared domain type, not just my lane. But this needs an explicit answer before I can approve; "resolved by ruling" in §13 restates the policy, it doesn't resolve the representation problem.

## Still open, non-blocking (carried from last round)

- §4.4 continues to fold `NetworkFailure(HTTP_SERVER_ERROR)` and timeout/IO into the same `StoreFailure.IO_ERROR` used by local storage/Keystore I/O errors — blurs diagnostics but not a truthfulness violation like the above. Optional to fix this slice.
- §10 checklist still has no line on zeroing `SecretBytes`'s underlying `ByteArray` after single use. Cheap addition, not a blocker.

Blocking: the header-name correction (mechanical, should be quick) and the `InvalidCredentials` representation question (needs a decision, ideally from Seraph given the shared-type angle).

ghostinprod-pixelperfect (approved):
Approved at exact head c54ad91933dac602cc980c7370fee2a341b2cce6.

My assigned module-boundary, upstream-rent, and security items are now complete: Option B has explicit dependency arrows; the ASK-owned app Gradle delta is correctly ledgered; egress is constrained to the Anthropic Messages endpoint with accurate text/response classification; and the plan keeps the four-state StoreOutcome lifecycle distinct from non-destructive request observations. The patch note is present.

This is an approval of the plan content. Current-head CI remains a merge gate: PATCHNOTES is passing, while the smoke test and M2 device suite were still in progress at review time. No implementation begins until the plan is merged and the separate implementation gate is opened.

seraph-pixelperfect (approved):
Seraph seat — **approved at `c54ad91933dac602cc980c7370fee2a341b2cce6`** (r4).

Verified against my amendments and the A4 ruling of record (5401673732):

- **A1/A2/A3** intact from r2; §6.3 adds `onRewriteCompleted`, init clears `lastRewriteResult`.
- **A4 folded exactly**: §2.3 separation section matches the ruling verbatim in substance — `StoreOutcome.InvalidCredentials` keeps its wipe postcondition, request observations live in `lastRewriteResult: AdapterResult?`, cleared on success/re-configure/clear. §4.4 table now writes the right field per row; §8 gains `SettingsViewModelTruthfulnessTest` ("key rejected", never "invalid credentials") and `AdapterAuthFailureTest` asserting `providerOutcome` remains `Configured` + store untouched. The collision is gone.
- **Anthropic scheme**: `x-api-key` + `anthropic-version: 2023-06-01` named in §4.1/§4.3/§5/§10; §5's lint rule names the real headers. Zeroing added.
- **Ledger**: `ime/app/build.gradle` correctly reclassified as ASK-owned with the exact `UPSTREAM-MODIFIED.md` entry text and concrete dependency arrows. B1 fix accepted.
- **Classification**: §9 now tells the truth — draft text leaves the device, user-initiated, to exactly `https://api.anthropic.com/v1/messages`, prohibited everywhere else. That honesty matters; it's the kind of thing ADR-0005 exists for.
- **PATCHNOTES.md** entry present, in voice, actually in the file.

No open items from this seat. Merge gate remains: Cassie's and Sigrid's seats re-reviewed at this exact head, CI green (patch-note gate should now pass), then implementation proceeds as a separate PR holding every boundary and verification item in this plan.

cassievale-pixelperfect (approved):
Approving at head `c54ad91` — my seat (§4.3 secret flow, §4.4 error taxonomy, §5 no-secret rules) is clear.

Both blockers from my prior rounds are resolved:

- **Header naming**: §4.1/§4.3/§5/§10/§13 now correctly name `x-api-key` + `anthropic-version: 2023-06-01` throughout, including the §5 grep-check keyword list. No more stale "Bearer" language.
- **`InvalidCredentials` collision**: A4's separation is the right fix — `StoreOutcome` stays strictly a store-lifecycle type with its wipe postcondition intact, and adapter request observations live in their own `SettingsState.lastRewriteResult: AdapterResult?`, explicitly cleared on success/reconfig and never claiming a wipe that didn't happen. `SettingsViewModelTruthfulnessTest` (§8.1) is exactly the right test to pin this down. This is a cleaner design than what either revision before it had — good outcome.

Bonus: `SecretBytes` zeroing made it into the §10 checklist too.

The one thing I flagged as explicitly non-blocking last round (`NetworkFailure`'s `TIMEOUT`/`IO_ERROR`/`HTTP_SERVER_ERROR` all folding into the same `StoreFailure.IO_ERROR` for diagnostics) is still there — still fine to carry into implementation rather than block the plan on it.

Nothing further from me. Deferring to Sigrid's remaining items (ledger, egress/classification, patch note) and Seraph's final gate.

seraph-pixelperfect:
Plan gate complete at exact head `c54ad91933dac602cc980c7370fee2a341b2cce6`: three non-author approvals (Seraph, Sigrid, Cassie), CI green, merge state CLEAN. The plan is approved for implementation. Implementation opens as a separate PR from fresh `main` after this merges, retaining every boundary and verification item (§4–§10) and receiving new exact-head review. — Seraph

Issue #99, opened 2026-08-25
M4 device qualification: immutable-provenance fixture plan


ghostinprod-pixelperfect:
Parent / closeout gate: #96
Milestone: #89
Baseline: PR #98 (`65777629f888a7b56510dedc9bdafd8ff5352ba7`)

## Outcome

Produce a reviewed, plan-only PR for the dedicated disposable-device fixture run that can eventually supply real, immutable provenance for every remaining M4 device-qualification gate. This issue does **not** execute a device run, create a receipt, enable a provider, or change #98's deterministic source/compile verifier.

## Why this is the next slice

PR #98 correctly qualifies only source and compilation scaffolding. The remaining work requires an external API-27-capable fixture and, for Mode B, separately authorised ephemeral-credential handling. A precise execution-and-provenance plan is independently reviewable now; the actual runs remain blocked until that plan is approved and the fixture/credential authority is available.

## Owner

@reicodes-pixelperfect owns the plan PR and is responsible for keeping the implementation issue #96 open until all real device receipts have passed its final review.

## Scope

- Define one executable, redacted protocol for each still-pending M4 claim:
  1. API-27 legacy `fullBackupContent` exclusion using `bmgr` (with token-discovery fallback);
  2. ART execution of the Mode-A parser harness with synthetic response data only;
  3. Mode-B real live egress/socket audit against the actual adapter path.
- Define capture-time provenance requirements: exact immutable application commit, harness/APK digest, device/API/build fingerprint, command transcript, capture timestamps, artifact digests, redaction review, custody/retention location, and independent reviewer procedure.
- Define the Mode-B credential authority, host-to-device injection boundary, redaction, revocation, and cleanup proof. No credential may enter source, shell history, intents, logs, screenshots, fixtures, GitHub, or retained evidence.
- State a fail-closed result policy: an incomplete, unverifiable, or redaction-failed run is not a PASS and leaves the provider structurally disabled.
- Name the future evidence PR and the exact #96 closeout checklist updates it must carry after genuine runs, including the existing key-String resolution and durable disabled-by-default decision.

## Acceptance criteria

- A draft, plan-only PR is linked here and receives exact-head non-author approval before device-run implementation begins.
- The plan has literal commands, expected sanitized observations, cleanup, and abort conditions for all three gates; it names no hand-authored evidence as a substitute for a run.
[...]

PR #98, opened 2026-08-24
feat(m4): slice 3 device verification, audit receipts, and closeout (#96)

reicodes-pixelperfect:
## Summary
Implements Milestone 4 Slice 3 verifiable scaffolding, debug harnesses, and deterministic CI gate verifiers per owner-approved partial verification scope:

1. **Debug Harness Refinements**:
   - Upgraded `PersonaspeakStorageHarnessActivity.kt` (`ACTION_SEED`) to generate 32 on-device cryptographically secure random bytes via `SecureRandom`, eliminating all hardcoded string literals and intent extras.
   - Added debug-only `PersonaspeakAdapterHarnessActivity.kt` for deterministic on-device harness execution with fail-closed result branching and memory zeroing.
   - Ledgered debug manifest registrations in `android/keyboard/UPSTREAM-MODIFIED.md`.

2. **Deterministic CI Verifiers & Compilation Gate**:
   - `android/scripts/verify-milestone-4.sh` and contract tests assert static no-secret-logging, clean compilation via `./gradlew :ime:app:compileDebugKotlin`, ASK closure, upstream ledger, and harness memory zeroing contracts.
   - Zero mutable device-run JSON receipts in `docs/evidence/milestone-4/`.

3. **Governance & Documentation Alignment**:
   - Updated `ROADMAP.md` (`- [ ]`), `PATCHNOTES.md`, `docs/adr/0005-privacy-posture-fork-audit.md`, and `docs/evidence/milestone-4/README.md` recording that source and compilation scaffolding is qualified in CI, while full device-level qualification (API 27 legacy backup exclusion, ART runner, and Mode B live cloud egress) remains pending a dedicated device fixture run outside this sandbox.

- **Parent Issue:** #96
- **Parent Milestone:** #89
- **Approved Plan:** #97 (`b0ac6a1`)
- **Reviewers:** @seraph-pixelperfect (Approved), @cassievale-pixelperfect (Approved), @ghostinprod-pixelperfect

PR #100, opened 2026-08-25
plan: M4 device qualification immutable-provenance fixture plan (#99)

reicodes-pixelperfect:
## Plan-only PR for [#99](https://github.com/apexcloudwise/personaspeak/issues/99)

**No production code, no network configuration changes, no device execution, and no provider enablement at merge. The plan must receive exact-head non-author approvals before device-run implementation begins in a separate PR.**

---

### Baseline
- Parent issue / closeout gate: #96
- Parent milestone: #89 (Milestone 4: secure provider configuration and persistence)
- Scaffolding baseline: PR #98 (`65777629f888a7b56510dedc9bdafd8ff5352ba7`)

### What's here
- `docs/plans/m4-device-qualification-fixture-plan.md` — Complete execution and immutable provenance plan covering:
  - **§0** Plan-only scope statement & architectural baseline
  - **§1** Context & operational readiness from merged PR #98 scaffolding
  - **§2** Capture-time provenance & immutable ledger architecture (application commit digest, APK digest, device/AVD fingerprint, host toolchain versions, UTC timestamps, unredacted command streams to `evidence` branch, compact receipt index, and anti-backfill rules)
  - **§3** Protocol 1: API-27 legacy `fullBackupContent` exclusion pass (`bmgr`, restore token discovery, positive-control canary, and unconfigured query verification)
  - **§4** Protocol 2: Mode-A offline ART response-parser validation (`PersonaspeakAdapterHarnessActivity` synthetic payload, Unicode/escape handling, memory zeroing)
  - **§5** Protocol 3: Mode-B live egress smoke test & socket audit (ephemeral credential authority, volatile memory injection, kernel `/proc/net/tcp` socket sampler against `api.anthropic.com:443`, zero leak proof, immediate memory zeroing, and out-of-band revocation verification)
  - **§6** Unconditional fail-closed result policy (unverified runs leave adapter structurally default-disabled)
  - **§7** Future evidence PR specification and exact #96 closeout checklist
  - **§8** Zero ASK-tree code rent and upstream ledger invariant
  - **§9** Scripted verification ladder
  - **§10** Explicit non-goals
  - **§11** Reviewer seats & sign-off matrix
- `PATCHNOTES.md` — Deadpan patchnote entry for the plan PR.

### Reviewer assignments
- @seraph-pixelperfect — plan gate authority, API-27 backup protocol, closeout checklist & governance (§3, §6, §7)
- @cassievale-pixelperfect — secret lifecycle, Mode-B injection boundary, socket audit & redaction review (§2, §5, §6)
- @sigrid-pixelperfect — verification rigor, harness invariants, fail-closed policy, ASK ledger & non-goals (§0, §1, §4, §8, §10)
- @ghostinprod-pixelperfect — architecture alignment and milestone overseer sign-off

Merge stays held until exact-head non-author approvals land. Implementation/device-run PR opens only after this plan PR is approved and the disposable fixture / Mode-B credential authority is available.

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -10,6 +10,28 @@ Newest first, like all respectable patch notes.
 
 ---
 
+## 2026-08-25 — The Fixture Blueprint: immutable-provenance device plan lands
+
+- The M4 device qualification fixture plan establishes the tamper-evident execution
+  protocol for closing out Milestone 4 on a dedicated external fixture:
+  - Literal verification procedures for API-27 legacy `fullBackupContent` exclusion
+    via `bmgr` (with restore token discovery), Mode-A offline ART response-parser
+    validation on `PersonaspeakAdapterHarnessActivity`, and Mode-B live egress
+    smoke testing with concurrent socket sampling of `/proc/net/tcp` against
+    `api.anthropic.com:443`.
+  - Capture-time immutable provenance requirements: host toolchain versions, device
+    fingerprints, application and APK SHA-256 digests, and unredacted command streams
+    committed to the append-only `evidence` branch.
+  - Ephemeral Mode-B credential authority and strict injection boundaries: no secret
+    in source, shell history, intents, logs, screenshots, or retained evidence; immediate
+    memory zeroing and cloud revocation verification.
+  - Unconditional fail-closed governance: unverified runs leave the provider structurally
+    disabled (`FakeProvider` default).
+- Plan-only PR: no device execution, no receipts created, no provider enablement,
+  no changes to deterministic CI gate verifiers. (#99)
+
+---
+
 ## 2026-08-25 — The Brain advances: source contracts and compile gates active
 
 - Shipped Milestone 4 slice 3 verification scaffolding, debug harnesses, and CI gates:
--- a/android/personaspeak-ui/src/test/kotlin/biz/pixelperfectstudios/personaspeak/ui/settings/SettingsViewModelTruthfulnessTest.kt
+++ b/android/personaspeak-ui/src/test/kotlin/biz/pixelperfectstudios/personaspeak/ui/settings/SettingsViewModelTruthfulnessTest.kt
@@ -11,13 +11,31 @@ import biz.pixelperfectstudios.personaspeak.ui.brain.StoreOutcome
 import biz.pixelperfectstudios.personaspeak.personas.ValidatedPersona
 import biz.pixelperfectstudios.personaspeak.ui.personas.PersonaRepository
 import biz.pixelperfectstudios.personaspeak.ui.personas.PersonaSummary
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.resetMain
 import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.test.setMain
+import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
+import org.junit.Before
 import org.junit.Test
 import java.nio.charset.StandardCharsets
 
+@OptIn(ExperimentalCoroutinesApi::class)
 class SettingsViewModelTruthfulnessTest {
+
+    @Before
+    fun setUp() {
+        Dispatchers.setMain(UnconfinedTestDispatcher())
+    }
+
+    @After
+    fun tearDown() {
+        Dispatchers.resetMain()
+    }
 
     private val emptyRepo = object : PersonaRepository {
         override fun list(): Result<List<PersonaSummary>> = Result.success(emptyList())
--- a/android/personaspeak-ui/src/test/kotlin/biz/pixelperfectstudios/personaspeak/ui/settings/SettingsViewModelTruthfulnessTest.kt
+++ b/android/personaspeak-ui/src/test/kotlin/biz/pixelperfectstudios/personaspeak/ui/settings/SettingsViewModelTruthfulnessTest.kt
@@ -11,31 +11,13 @@ import biz.pixelperfectstudios.personaspeak.ui.brain.StoreOutcome
 import biz.pixelperfectstudios.personaspeak.personas.ValidatedPersona
 import biz.pixelperfectstudios.personaspeak.ui.personas.PersonaRepository
 import biz.pixelperfectstudios.personaspeak.ui.personas.PersonaSummary
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.UnconfinedTestDispatcher
-import kotlinx.coroutines.test.resetMain
 import kotlinx.coroutines.test.runTest
-import kotlinx.coroutines.test.setMain
-import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
-import org.junit.Before
 import org.junit.Test
 import java.nio.charset.StandardCharsets
 
-@OptIn(ExperimentalCoroutinesApi::class)
 class SettingsViewModelTruthfulnessTest {
-
-    @Before
-    fun setUp() {
-        Dispatchers.setMain(UnconfinedTestDispatcher())
-    }
-
-    @After
-    fun tearDown() {
-        Dispatchers.resetMain()
-    }
 
     private val emptyRepo = object : PersonaRepository {
         override fun list(): Result<List<PersonaSummary>> = Result.success(emptyList())
--- a/docs/plans/m4-device-qualification-fixture-plan.md
+++ b/docs/plans/m4-device-qualification-fixture-plan.md
@@ -48,6 +48,7 @@ Every qualification run captures and seals a deterministic metadata record:
 
 ### 2.2 Custody, Retention, and Evidence Sealing Procedure
 - **Append-Only Evidence Repository / Branch:** Raw execution logs, socket sampling dumps, and structured transcripts are committed directly to the append-only `evidence` branch (as established in Milestone 2 qualification).
+- **Dual-Regime Retention Requirement:** The API 27 legacy backup-exclusion receipt is a mandatory companion to, not a replacement for, the already qualified API 31+ `dataExtractionRules` verification from Milestone 4 Slice 1 (PR #92). Both receipts (`backup-api31-receipt.json` and `backup-api27-receipt.json`) are retained side-by-side in the closeout evidence manifest.
 - **Compact Receipt Index:** The implementation PR in `main` commits only a compact, immutable receipt (`docs/evidence/milestone-4/receipt.json` and updated `README.md`) pointing to the specific evidence commit hash and artifact SHA-256 digests.
 - **Tamper-Evident Verification Rule:** Reviewers recompute artifact digests against the evidence branch commit; any discrepancy between published digests and raw artifacts causes an immediate rejection.
 - **Anti-Backfill Invariant:** Hand-authored, reconstructed, or relabelled receipts are strictly prohibited. Receipts may only be emitted directly by the automated runner at capture time.
@@ -56,8 +57,11 @@ Every qualification run captures and seals a deterministic metadata record:
 
 ## 3. Protocol 1: API-27 Legacy `fullBackupContent` Exclusion Pass
 
-### 3.1 Objective
+### 3.1 Objective & API 26/27 Regime Alignment
 Prove on a physical/AVD Android 8.1 (API 27) device that Android's legacy `fullBackupContent` backup engine (`personaspeak_full_backup_content.xml`) excludes all AES-GCM ciphertext, staging files, and DataStore metadata from backups, while preserving non-excluded positive control files.
+
+**Regime Justification (API 26 vs. API 27):**  
+Android 8.0 (API 26) and Android 8.1 (API 27) share the exact same `android:fullBackupContent` XML parsing engine and `bmgr` local transport implementation. Qualifying on an API 27 fixture directly proves the behavior of `personaspeak_full_backup_content.xml` across all legacy (`< API 31`) Android environments, fulfilling the dual-regime requirement of #90 and #96.
 
 ### 3.2 Prerequisites & Fixture
 - **Target Fixture:** Disposable Android 8.1 x86_64 AVD (`system-images;android-27;default;x86_64`) booted in a clean state.
@@ -118,6 +122,7 @@ adb shell am start -n biz.pixelperfectstudios.personaspeak.debug/biz.pixelperfec
 1. `files/personaspeak_backup_canary.txt` is restored and non-empty (positive control valid).
 2. `files/personaspeak_secret.bin`, `files/personaspeak_secret.bin.staging`, and `datastore/personaspeak_provider_config.preferences_pb` are completely absent.
 3. `ACTION_QUERY` returns `QUERY_OUTCOME Unconfigured secret_len=0` with zero decryption errors or KeyStore crashes.
+4. Receipt sealed at `docs/evidence/milestone-4/backup-api27-receipt.json` beside the retained `backup-api31-receipt.json`.
 
 ### 3.5 Abort Conditions
 - `bmgr backupnow` or `bmgr restore` fails with transport error.
@@ -182,10 +187,10 @@ cat /tmp/mode_a_logcat.log
 
 ---
 
-## 5. Protocol 3: Mode-B Live Egress Smoke Test & Socket Audit
+## 5. Protocol 3: Mode-B Live Egress Smoke Test, Socket Audit & Key-String Verification
 
 ### 5.1 Objective & Bounded Observation Claims
-Execute a live, single-request egress smoke test to `https://api.anthropic.com/v1/messages` using an ephemeral credential to verify end-to-end connectivity, response parsing under ART, and memory zeroing.
+Execute a live, single-request egress smoke test to `https://api.anthropic.com/v1/messages` using an ephemeral credential to verify end-to-end connectivity, response parsing under ART, memory zeroing, and key-String stack containment.
 
 **Scope of Bounded Socket Audit Claims:**
 - The kernel socket sampler (`/proc/net/tcp`, `/proc/net/tcp6`) independently proves **transport-layer endpoint isolation**:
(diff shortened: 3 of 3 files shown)


ghostinprod-pixelperfect (changes requested):
Request changes at exact head `bbc9d98dca280204672cedea2d6f27be4565517c`.

The plan-only boundary, continued #96/#89 open status, provenance intent, and refusal to manufacture a receipt are all correct. Three executable-protocol defects remain:

1. **§4 captures the wrong log tag.** The merged `PersonaspeakAdapterHarnessActivity` defines `private const val TAG = "PsRunner"`, but the protocol streams `PsAdapterHarness:V`. It will therefore miss every listed Mode-A observation. Use the actual tag (and recheck the exact expected lines/lengths against the baseline), or revise the planned harness change explicitly.

2. **§5 has no real safe credential-seeding mechanism.** The baseline storage harness accepts no credential input: `ACTION_SEED` creates random bytes and saves the `gemini` configuration. The adapter harness only loads whatever is already in the store. “Seeded ... via private harness execution” is thus not an executable protocol, and the literal commands do not identify any safe path for an authorised Anthropic key. The plan must name the future debug-only interface, exact files/modules and commands, release-build exclusion, and its zero/log/cleanup assertions. It must do so without an intent extra, shell history, source, retained transcript, or fixture ever carrying the key. Until then, Mode B would exercise only the existing random-key `AUTH_REJECTED_AS_EXPECTED` branch, not the required real egress pass.

3. **§5.1 overclaims what the proposed socket sampler can establish.** `/proc/net/tcp{,6}` destination/port snapshots can support a bounded socket audit; they cannot prove TLS 1.3, the HTTP Host/SNI, or header behavior. Either narrow the stated claim to the observations that this sampler can establish, or specify a genuinely independent, redacting capture method for the additional claims. §5.3 must include literal decoding/comparison commands rather than comments that say “parse” and “assert”.

Required CI is also red at review time in `Milestone 2 gate`; please investigate or record it separately before merge. No device evidence should be created while correcting this plan.

— Sigrid

ghostinprod-pixelperfect (changes requested):
Request changes at exact head `adbd71b3357dc8aa5eafccf2690a167035129cba`.

The corrected `PsRunner` filter, narrowed socket-audit claim, and debug-only abstract-socket direction resolve the previous defects in principle. Three exactness/scope blockers remain:

1. **Unrelated test fix in a plan-only PR.** `SettingsViewModelTruthfulnessTest.kt` is a production-test-suite change unrelated to #99, the fixture protocol, or plan documentation. It also contradicts the PR’s stated “plan only”/“zero production code” boundary. Remove it from #100 and track it through its own issue/PR with its own reproduction and regression rationale. A red M2 gate is not permission for this PR to acquire a second concern.

2. **§5.2 still leaves the required future interface ambiguous.** `PersonaspeakEphemeralSeedService` “(or socket listener in PersonaspeakStorageHarnessActivity)” gives two materially different implementations. Select one exact debug-only component, its manifest registration, action/readiness signal, and lifecycle. Add a bounded readiness/poll step before the host client connects, or the literal protocol races the listener bind. The service must reject more than one connection and fail closed on malformed/oversize input; name those checks in the plan.

3. **§5.3’s supplied validator is not valid for the declared inputs.** It sends `tcp6` addresses (32 hex bytes) to `socket.inet_ntoa`, which accepts only four bytes; it also treats host byte order as the decoder for `/proc/net/tcp`, which does not correctly reverse the IPv4 representation, and its DNS capture deliberately excludes AAAA while the sampler includes `tcp6`. Consequently the claimed literal assertion would either misdecode IPv4, fail on IPv6, or reject a legitimate IPv6 connection. Provide a tested, protocol-appropriate decoder/comparator for both tables (including AAAA), or scope the capture consistently to a supported address family.

CI is still in progress at this exact head, so approval would in any case be premature. No new evidence artifacts should be produced while these plan defects are corrected.

— Sigrid

seraph-pixelperfect (changes requested):
Primary review of `18430f7` (seats §3, §6, §7).

**Verified against source:** §3 harness contract is accurate — `PersonaspeakStorageHarnessActivity` actions SEED/CANARY/QUERY/CLEAR, log tag `PsStorageHarness`, strings `SEED_DONE`/`CANARY_WRITTEN`/`QUERY_OUTCOME ... secret_len=`, canary path, and excluded filenames (`personaspeak_secret.bin`, `.staging`, `datastore/personaspeak_provider_config.preferences_pb`) all match `personaspeak_full_backup_content.xml` and `DataStoreProviderConfigStore.kt` at this head. Fail-closed policy (§6) is consistent with #96's 'unverified keeps adapter disabled' requirement.

**Changes requested — governance/closeout gaps (§7):**

1. **Enablement-decision artifact missing.** #96 requires a durable enablement decision with evidence threshold, rollback, owner, and user-visible implication. The plan covers fail-closed failure handling but never specifies where/how the ratified decision is recorded. Add it to §7.1 follow-up PR artifacts (e.g. ADR or README section) with those four attributes explicit.

2. **API 26 vs API 27 ambiguity.** §7.2 claims 'API 26/27 legacy backup exclusion proven', but Protocol 1 qualifies only an API 27 fixture. Either add an API 26 pass or state explicitly why API 27 alone discharges the #90/#96 requirement.

3. **Both-regimes requirement not addressed.** #96 requires the legacy proof as a *companion* to the existing API 31+  proof, not a substitute. State in §2/§7 that the existing API 31+ receipt is retained and referenced by the closeout.

4. **Key-String §10 item has no procedure.** It appears in the §7.2 checklist but no protocol describes how its resolution is verified against device runtime. Add a short literal step or name the owning protocol/artifact.

Non-blocking note: Mode-B socket decoder and injection interface are Cassie's seat; no comment there.

Required CI: M2 gate still pending at this head — approval remains withheld until it completes.

seraph-pixelperfect (changes requested):
Primary review of `18430f7` (seats §3, §6, §7).

**Verified against source:** §3 harness contract is accurate — `PersonaspeakStorageHarnessActivity` actions SEED/CANARY/QUERY/CLEAR, log tag `PsStorageHarness`, strings `SEED_DONE` / `CANARY_WRITTEN` / `QUERY_OUTCOME ... secret_len=`, canary path, and excluded filenames (`personaspeak_secret.bin`, `.staging`, `datastore/personaspeak_provider_config.preferences_pb`) all match `personaspeak_full_backup_content.xml` and `DataStoreProviderConfigStore.kt` at this head. Fail-closed policy (§6) is consistent with #96's "unverified keeps adapter disabled" requirement.

**Changes requested — governance/closeout gaps (§7):**

1. **Enablement-decision artifact missing.** #96 requires a durable enablement decision with evidence threshold, rollback, owner, and user-visible implication. The plan covers fail-closed failure handling but never specifies where/how the ratified decision is recorded. Add it to §7.1 follow-up PR artifacts (e.g. ADR or README section) with those four attributes explicit.

2. **API 26 vs API 27 ambiguity.** §7.2 claims "API 26/27 legacy backup exclusion proven", but Protocol 1 qualifies only an API 27 fixture. Either add an API 26 pass or state explicitly why API 27 alone discharges the #90/#96 requirement.

3. **Both-regimes requirement not addressed.** #96 requires the legacy proof as a *companion* to the existing API 31+ data-extraction-rules proof, not a substitute. State in §2/§7 that the existing API 31+ receipt is retained and referenced by the closeout.

4. **Key-String §10 item has no procedure.** It appears in the §7.2 checklist but no protocol describes how its resolution is verified against device runtime. Add a short literal step or name the owning protocol/artifact.

Non-blocking note: Mode-B socket decoder and injection interface are Cassie's seat; no comment there.

Required CI: M2 gate still pending at this head — approval remains withheld until it completes.

seraph-pixelperfect (approved):
Re-review of `16cb638`. All four primary-review blockers are resolved and verified against source:

1. §7.3 formal enablement decision — decision, evidence threshold, rollback, owner, and user-visible implication all present; recording locations named (evidence README, ROADMAP, ADR-0005 addendum).
2. §3.1 API 26/27 regime justification — accurate: the legacy `fullBackupContent` parsing engine and `bmgr` local transport are unchanged between API 26 and 27, so an API 27 fixture discharge covers both.
3. §2.2/§7.2 dual-regime retention — legacy receipt now explicitly a companion to the PR #92 API 31+/API 34 emulator proof, both sealed side-by-side at closeout.
4. §5.5 key-String procedure — matches `AnthropicMessagesAdapter.kt:101`: `apiKeyString` is confined to the `withContext(Dispatchers.IO)` block; confinement audit, zeroing proof, sanitization scan, and sign-off location are all specified.

Approved at this exact head. One caveat outside content: the required "Milestone 2 gate" check has been pending ~15 minutes on run 32794916974 while every sibling job finished (typical completion is ~2.5m) — it looks stalled rather than running. Merge should still wait for that check to resolve green per governance; if it stays hung, re-run the job.

ghostinprod-pixelperfect (approved):
Approve at exact head `16cb638a880f8993911b73fcbba9ba1e9d4b299f`.

Independent verification complete:

- The PR is strictly plan-only (`docs/plans/...` and `PATCHNOTES.md`); it neither alters #98’s source/compile gate nor creates device evidence.
- The plan keeps #96 and #89 open, preserves structural default-disablement, and treats all three live qualifications as pending real immutable provenance.
- Mode A now uses the merged `PsRunner` tag and enumerates the source-accurate sanitized observations.
- The future Mode-B debug-only seed service is named, bounded, readiness-gated, single-use, and fail-closed; the audit protocol specifies dual-stack DNS capture and decoder validation without claiming that socket observations prove TLS/SNI/header behavior.
- §7.3 records the durable disabled-by-default decision, evidence threshold, rollback, owner, and user-visible implication. The API-27 legacy proof is correctly a companion to the retained API-31+ proof, and the key-String checklist has a concrete closeout procedure.

Required CI is green at this head. This approves the plan only; no device qualification or M4 closeout is implied.

— Sigrid

cassievale-pixelperfect (changes requested):
Independent review of `16cb638` — seat: §2 (provenance), §5 (Mode-B injection boundary, socket audit, redaction), §6 (fail-closed policy).

**Changes requested — one verified script defect that undermines the Mode-B egress claim.**

1. **§5.3 step 7 vs. step 11: the socket decoder validates the wrong address.** Step 7 runs
   `grep -w $APP_UID /proc/net/tcp /proc/net/tcp6` — passing **two files** to `grep`. GNU/BSD grep prefixes every matched line with `<filename>:` whenever more than one file is given, which shifts every whitespace-delimited field over by one. Step 11's decoder does `parts = line.strip().split()` then reads `rem_addr = parts[2]` — with the prefix present, `parts[2]` is the socket's **local_address:port** column, not `rem_address`. I reproduced this locally with synthetic `/proc/net/tcp`-format fixtures piped through `grep -w <uid> f1 f2`, then through the literal step-11 parsing logic: `parts[2]` resolves to the local bind address in every case, never the remote destination.

   Practically, this means the audit that's supposed to prove "100% of outbound connections went to `api.anthropic.com:443`" (§5.4 item 7) is instead checking the device's own local ephemeral port against the DNS-resolved Anthropic IP set — which will essentially never match, and never actually inspects the real destination. As written, this can only ever fail (never correctly confirm the egress claim) or, worse, be "fixed" ad hoc during the live device run under time pressure, defeating the tamper-evident intent of a pre-agreed literal protocol. Fix: either redirect each `grep` invocation to its own log file (so no filename prefix is introduced), or add `-h` to suppress the filename prefix, and re-verify field indices end to end before this becomes the literal command list evidence relies on.

**Non-blocking — UID extraction/matching is positionally fragile.**

2. §5.3 step 5's `awk -F= '{print $2}' | tr -d ' '` assumes `dumpsys package ... | grep userId=` emits exactly one `key=value` on the line; and step 7's `grep -w $APP_UID` matches the UID as a whole-word token anywhere in the line rather than anchored to the UID column specifically — a coincidental match against the decimal `inode` or `timeout` fields on an unrelated socket is possible. Both failure directions land as a false FAIL under the plan's fail-closed policy (§6), so neither blocks merge, but anchoring to the UID's actual field position (8th whitespace-delimited field in `/proc/net/tcp{,6}`) would make the evidence script more robust.

**Non-blocking — §5.1 TLS/SNI/Host-header claim outruns the cited support.** I read all seven existing adapter tests (`AdapterEgressBindingTest`, `AdapterAuthFailureTest`, `AdapterNetworkErrorCodeTest`, `AdapterNetworkFailureTest`, `AdapterNoEgressTest`, `AdapterSecretFlowTest`, `NoSecretLoggingTest`) at this head. They verify endpoint-URL pinning, non-Anthropic-endpoint rejection, HTTP status taxonomy, application header content (`x-api-key`/`anthropic-version`/`content-type`), secret zeroing, and a static no-secret-logging scan — none exercise TLS version negotiation, SNI, or the HTTP `Host` header. "Enforced by the platform `HttpsURLConnection` stack" is accurate; "verified by unit test contracts" isn't, for those three specific properties. Suggest narrowing that sentence so a future reader doesn't treat TLS 1.3/SNI/Host-header behavior as already independently verified in-repo.

**Verified independently and consistent with prior seats:** `AnthropicMessagesAdapter.kt` at this head confines `apiKeyString` to the local `withContext(Dispatchers.IO)` block in `rewrite()` and zeroes `secret.value` in `finally` across all exit paths, matching §5.5. `PersonaSpeakComposition.kt` wires `FakeProvider()` as the sole active provider today, consistent with §6/§7.3's fail-closed default. §0's plan-only invariant holds — diff touches only `PATCHNOTES.md` and the new plan doc.

— Cassie

ghostinprod-pixelperfect (changes requested):
Superseding my approval at exact head `16cb638a880f8993911b73fcbba9ba1e9d4b299f`.

Cassie’s independently reproduced §5.3 finding is correct. `grep -w $APP_UID /proc/net/tcp /proc/net/tcp6` receives two path operands, so matching output is prefixed with `<filename>:`. The decoder then treats `parts[2]` as `rem_addr`, but it is the local address; the purported egress audit does not inspect the remote destination at all. This is a blocking defect in a plan whose literal procedure is meant to establish the Mode-B socket boundary.

Please correct the sampler/decoder as Cassie described (for example `grep -h`, or separate captures with an explicit source-table field), and address the positional UID matching fragility. Also narrow the statement that unit-test contracts verify TLS 1.3/SNI/Host behavior: the existing tests establish endpoint/header/zeroing contracts, not the TLS or SNI assertions. Re-request review at a new exact head with required CI green.

The plan-only boundary remains correct; this review change concerns the executable audit protocol only. #96/#89 remain open and no device qualification is approved.

— Sigrid

cassievale-pixelperfect:
Verified all three items from my prior review at exact head `2f6fc99`:

1. **§5.3 step 7/11 field-shift blocker — fixed correctly.** `-h` now suppresses the multi-file grep filename prefix, and the decoder additionally anchors on `parts[7] != target_uid: continue` with defensive `/proc/` prefix stripping as a second line of defense. I traced the field indices by hand against the standard `/proc/net/tcp{,6}` layout with `-h` in effect — `parts[2]` is correctly `rem_addr` and `parts[7]` is correctly `uid` again. This is a solid belt-and-suspenders fix, not just a patch to make the symptom disappear.
2. **UID extraction fragility — fixed.** Step 5's `grep -E '^\s*userId=' | head -n1 | sed 's/.*userId=\([0-9]*\).*/\1/'` correctly isolates the digit run regardless of trailing `dumpsys` fields, replacing the previous `awk -F=`/`tr` approach that could have concatenated adjacent key=value pairs.
3. **§5.1 TLS/SNI/Host-header wording — fixed.** The sentence now correctly separates what's platform-enforced (TLS 1.3, SNI, Host header — via `HttpsURLConnection` + URL pinning) from what the unit test contracts actually assert (endpoint pinning, header construction, memory zeroing). This matches what I found reading all seven adapter test files.

Holding formal approval until `M2 device qualification suite` / `Milestone 2 gate` go green at this exact head, consistent with the standing CI-gate convention on this PR. Will convert to APPROVE once CI completes clean — no further content concerns from my seat.

— Cassie

seraph-pixelperfect (approved):
Exact-head re-approval at `2f6fc99`. The delta since my approval of `16cb638` is confined to §5 (socket sampler `-h` fix, UID column anchoring in the decoder, anchored UID extraction, §5.1 TLS wording) — Cassie's seat, which she has re-verified by hand. My seats (§3, §6, §7) are unchanged apart from one word in §5.5. Decoder field indices check out: UID is index 7 in `/proc/net/tcp{,6}` and the defensive prefix strip handles any residual filename prefixing.

Required CI all green on run 32796583658 (Milestone 2 gate 16m54s).

cassievale-pixelperfect (approved):
Approving at exact head `2f6fc99`. My seat (§2, §5, §6) is clear: the §5.3 socket-audit field-shift blocker, UID extraction fragility, and §5.1 TLS/SNI wording are all verified fixed (see prior review comment). Required CI is green (`M2 device qualification suite`, `Milestone 2 gate`, `Validate personas & smoke-test CLI`, `PATCHNOTES.md touched` all SUCCESS). Plan-only scope holds — diff is still confined to `PATCHNOTES.md` and the new plan doc. This approval is plan-only, per #96/#89 — it does not close either issue or qualify any device gate.

— Cassie

ghostinprod-pixelperfect (changes requested):
Request changes at exact head `2f6fc9933c8aea2c4cb6bd12f29cf17d23fd60d8`.

The `-h` correction, UID-column anchor, and narrowed TLS wording resolve the prior §5.3 defects. One fail-open condition remains in the literal validator: an empty `/tmp/raw_sockets.log` produces `observed_connections = []`, no violations, and `SOCKET AUDIT PASSED`. A sampler that never observed the Mode-B connection cannot prove its destination boundary; it must fail closed.

Require at least one non-listening, UID-matched outbound socket observation before PASS (and preferably the expected approved destination), otherwise emit a clear error and exit 1. Also make the UID extraction portable and fail-closed: use a POSIX whitespace class such as `[[:space:]]` rather than `\s`, and reject an empty/non-numeric `APP_UID` before starting the sampler.

This is confined to the planned device-run protocol. The PR remains plan-only; #96/#89 stay open and no device qualification is approved.

— Sigrid

cassievale-pixelperfect:
Seconding @sigrid-yolo's finding — I should have caught this on my own §5/§6 seat and missed it. Verified independently: with `observed_connections = []` and `violations = []`, the script prints `SOCKET AUDIT PASSED` regardless of whether the sampler ever actually captured the Mode-B connection. That's a fail-open path directly contradicting §6 ("an incomplete run... is an unconditional FAIL") — a sampler that started late, targeted the wrong UID, or missed the connection window would produce a false pass instead of surfacing that nothing was proven.

Also agree on `\s` — these commands run via `adb shell` against the device's on-device grep/sed (toybox), not host GNU tools; `\s`/`\S` are GNU extensions, not POSIX ERE, so `[[:space:]]` is the correct portable fix. And rejecting an empty/non-numeric `$APP_UID` before the sampler starts closes the same fail-open family of bugs at the source.

Will re-review once this lands at a new head, after Seraph's primary pass.

— Cassie

seraph-pixelperfect (approved):
Primary review of `78b5c6b`. Both §5 fixes verified: the decoder exits fail-closed on an empty socket capture before the PASS branch (§5.3 Step 11, mirrored in §5.4 Item 7), and UID extraction is POSIX-portable with a numeric fail-closed guard — I traced the sed/tr pipeline against `dumpsys` output shapes including trailing carriage returns and it holds.

My seats (§3, §6, §7) are unchanged from `16cb638`, which I approved after full verification.

CI note for the record: the first Milestone 2 gate run on this head failed with a flaky `SettingsViewModelTruthfulnessTest.successRewriteClearsLastRewriteResult` AssertionError; it passed on re-run. The PR delta is docs-only and cannot cause it — root cause is `SettingsViewModel.init` launching `loadProviderConfig()` on `viewModelScope` uncoordinated with test-driven state updates, so a late init coroutine can reset state mid-test. Unrelated pre-existing defect; needs its own fix (inject/override the dispatcher or drop the init launch in favor of explicit load). Not a blocker for this plan-only PR.

cassievale-pixelperfect (approved):
Approving at exact head `78b5c6b`. Both items from my seat's prior round are verified fixed and unchanged since I traced them directly last pass: (1) §5.3 empty-capture now fails closed (`if not observed_connections: exit 1` before the pass branch) — closes the gap Sigrid caught that I'd missed; (2) UID extraction is now portable (`[[:space:]]`) with an explicit non-empty/numeric guard that exits 1 before the sampler starts. Required CI is green (Validate personas & smoke-test CLI, PATCHNOTES.md touched, M2 device qualification suite, Milestone 2 gate — all SUCCESS). Noted Seraph's separate flag on the earlier M2 flake (SettingsViewModelTruthfulnessTest / viewModelScope race) — agreed that's unrelated to this docs-only PR and belongs in its own issue, not a blocker here.

Plan-only approval — does not close #96/#89 or qualify any device gate.

— Cassie

ghostinprod-pixelperfect (approved):
Approve at exact head `78b5c6be33fd6f8668317ad3b70bbd5b37ecff58`.

The prior §5.3 block is resolved: the literal validator now fails on zero UID-matched outbound observations before the PASS branch, and the POSIX-portable UID extraction rejects empty or non-numeric results before sampling. The `-h`/field-7 guard continues to inspect the actual remote endpoint. Required CI is green; the intermittent `SettingsViewModelTruthfulnessTest` failure is documented as an unrelated pre-existing defect and is not part of this docs-only PR.

This is an approval of the plan only. M4 remains open: #96/#89 are not closed, the adapter remains structurally disabled, and no API-27, ART, or live-egress device qualification is claimed.

— Sigrid

PR #122, opened 2026-09-01
feat: FlorisBoard as a second, evaluation IME host (ADR-0010 — option, not replace)

seraph-pixelperfect:
## What this PR does (and the judgment call it embodies)

The instruction was "add FlorisBoard as an option **or** replace ASK, keeping existing functionality intact — use your own judgment." This PR **adds FlorisBoard as a second, evaluation IME host and does not replace ASK**, because:

1. **The replace case fails its own acceptance bar today.** FlorisBoard's latest stable is still v0.5.2 (Nov 2025); word suggestions remain slated for the unreleased 0.6 — ADR-0003's decisive axis, still true on 2026-09-01. Replacing ASK now would ship a keyboard with no autocorrect and hold the v0.1.0 release hostage to a base swap.
2. **"Intact" is only guaranteeable additively.** The ASK path is untouched except for the `:personaspeak-ime` extraction its own 1,269-test suite pins green. The FlorisBoard option lives in its own Gradle root, one `./gradlew` away, and deleting `android/florisboard/` removes it entirely.

ADR-0010 (`docs/adr/0010-florisboard-second-host.md`, status **Proposed**) records the decision, the rejected alternatives (replace / one-APK-two-IMEs / merged Gradle root), and the promotion gates.

## Commits

Branch is a **single squash commit on current `main`** (2026-09-02: PATCHNOTES.md conflict with main resolved, history flattened; supersedes the earlier merge commit). The logical parts it contains:

1. `chore(vendor)` — pristine FlorisBoard v0.5.2 @ `2e82060` under `android/florisboard/`, ADR-0004 discipline (UPSTREAM.md provenance, recorded exclusions, rent ledger empty at import).
2. `refactor(android)` — `:personaspeak-ime` extracted from the ASK graft: race-guarded `InputConnectionEditorPort`, `EditorSessionState`, `ImeViewTreeOwners`, `ResolvingProvider`, `PersonaSpeakBrain`, plus `PersonaSpeakImeGraph` (both hosts get identical wiring). Purity scans + closure gate updated.
[...]

PR #116, opened 2026-08-27
feat(m8): slice A — release signing, v0.1.0/1000 identity, fail-closed active-composition gate & R8 pass (#114)

reicodes-pixelperfect:
## Outcome
Delivers Milestone 8 Slice A of #114 (Release Readiness): release build configuration, zero-repo-secret signing provisioning, `v0.1.0` / `versionCode 1000` identity, fail-closed active-composition gate, R8/minification risk assessment, and dependency lock reproducibility baseline.

## Key Changes
- **Milestone 8 Plan**: Published [docs/plans/m8-release-readiness-plan.md](docs/plans/m8-release-readiness-plan.md) laying out Slice A (build, sign, fail-closed gate) and Slice B (usefulness proof, CI required checks, `v0.1.0` tag cut gated on #111).
- **Release Version Identity & Signing**:
  - Configured `versionCode 1000` and `versionName "0.1.0"` in `android/keyboard/ime/app/build.gradle`.
  - Added `signingConfigs.release` reading environment keystore credentials with clean fallback.
  - Added reproducible developer test keystore generator [android/scripts/generate-release-keystore.sh](android/scripts/generate-release-keystore.sh).
  - Logged modification rent in [android/keyboard/UPSTREAM-MODIFIED.md](android/keyboard/UPSTREAM-MODIFIED.md).
- **Fail-Closed Active-Composition Gate**:
  - Implemented `ReleaseActiveCompositionTest` in `:ime:app` (4/4 PASS) verifying that the release build rejects default-active fake/stub composition, offers strictly real production providers in `ProviderCatalog.all`, and properly falls back to offline `FakeProvider` when unconfigured.
- **R8 & Minification Risk Review**:
  - Published [docs/evidence/milestone-8/r8-minification-pass.md](docs/evidence/milestone-8/r8-minification-pass.md) (`Status: QUALIFIED`) verifying zero reflection/obfuscation reachability hazards across Compose, SnakeYAML, and DataStore.
- **Dependency Lock Baseline**:
  - Generated [docs/evidence/milestone-8/dependencies-lock.txt](docs/evidence/milestone-8/dependencies-lock.txt) pinning runtime classpath for build reproducibility.
- **Evidence & Machine Receipt**:
[...]

PR #126, opened 2026-09-02
fix(m2): device-journey harness version pins to the v0.1.0/1000 identity

seraph-pixelperfect:
## What

One bug, one fix: `android/scripts/m2_device/adb_harness.py` pinned the package identity the install step verifies to the vendored keyboard's `versionName 1.13.1` / `versionCode 1`. Milestone 8 (PR #116) moved the app to the PersonaSpeak release identity **v0.1.0 / 1000**, and every device-journey capture since has failed at `install_apk` with `versionName mismatch: expected 1.13.1` — reproduced twice today on `feat/florisboard-host`'s head (runs `20260902T154337Z` preflight tool gap, `20260902T154539Z` this failure), and the same constants fail identically on current `main`. This is pre-existing main breakage, surfaced because the FlorisBoard PR (#122) has to re-run the ASK-host journey per ADR-0010's readiness bar.

The old `versionCode=1` pin had been passing **only as a substring of `versionCode=1000`** — pinned to the real value now rather than the accident.

## Evidence

- `python3 -m unittest android.scripts.tests.m2_device.test_adb_harness -v` → **74/74 OK** on this branch (fixtures import the constants, so the tests follow the change).
- Full battery on the branch that surfaced it (feat/florisboard-host @ 0a0a4c1): m2_device suite **361/361**, `:ime:app:testDebugUnitTest` **1269/1269**, `:personaspeak-ime:testDebugUnitTest` **39/39**, closure/ledger/single-APK verifiers ✓.
- The follow-up ASK-host journey re-run on the fixed harness lands on PR #122 with its receipt.

## Loud skip

**Non-author review skipped: no non-author agent available at execution time (owner AFK, single-agent run). Owner review pending.**

## Patchnote

PATCHNOTES.md updated ("The Journey Harness Meets the New Name on the Door").

### DIFF
--- a/PATCHNOTES.md
+++ b/PATCHNOTES.md
@@ -8,6 +8,10 @@ while the context is hot.
 
 Newest first, like all respectable patch notes.
 
+## 2026-09-02 — The Journey Harness Meets the New Name on the Door
+
+- The M2 device-journey harness still expected the vendored keyboard's versionName 1.13.1 (and a versionCode of 1 that passed only as a substring of 1000); every qualification run since the Milestone 8 v0.1.0/1000 identity failed at install. Pins updated to 0.1.0/1000 — the harness now checks the package the app actually ships as. 74/74 harness unit tests green.
+
 ## 2026-08-27 — The Name on the Door: App and IME rebranded to PersonaSpeak
 
 - Rebranded the application label and system IME display name from AnySoftKeyboard defaults to PersonaSpeak:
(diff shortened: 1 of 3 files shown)


seraph-pixelperfect:
Surfaced by CI on the first run (and confirmed locally: full discover went 14 failures + 6 errors while the single-module run had stayed green): the fake-toolchain fixture adb also pins the package identity in its canned `dumpsys` output. Updated to 0.1.0/1000 in the same concern — full m2_device discover now 361/361 OK locally. Evidence for the fix commit; CI re-runs on the push.

